PR #7, opened 2018-10-13
Update README.md

Tylertron1998:
Very better grammar with many nicer change.


kyranet (approved):
Sexy grammar, total 100/100. Thanks for this!

PR #161, opened 2019-05-27
Connect Four rewrite

kyranet:


kyranet:
The AI will be fixed later, right now it's priority to get the Human-vs-Human version to work.

PR #168, opened 2019-05-30
Using dev config for lavalink setup

Skillz4Killz:
This PR disables lavalink setup to run a development bot.

- [x] Creates a `DEV_LAVALINK` constant that will be used to enable/disable lavalink.
- [x] By default the config.example will have this set to true as most developers will need it for development

### DIFF
--- a/config.ts.example
+++ b/config.ts.example
@@ -5,6 +5,7 @@ import { RPoolConnectionOptions } from 'rethinkdb-ts';
 import { APIWebhookData } from './src/lib/types/DiscordAPI';
 
 const DEV = 'DEV' in process.env ? process.env.DEV === 'true' : !('PM2_HOME' in process.env);
+export const DEV_LAVALINK = true
 
 export const DATABASE_DEVELOPMENT: RPoolConnectionOptions = { db: 'test' };
 export const DATABASE_PRODUCTION: RPoolConnectionOptions = {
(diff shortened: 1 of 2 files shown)


src/lib/SkyraClient.ts
```diff
@@ -58,14 +58,16 @@ export class SkyraClient extends KlasaClient {
...
 	@enumerable(false)
-	public lavalink: Lavalink = new Lavalink({
-		send: async (guildID: string, packet: any) => {
-			const guild = this.guilds.get(guildID);
-			if (guild) this.ws.shards.get(guild.shardID).send(packet);
-			else throw new Error('attempted to send a packet on the wrong shard');
-		},
-		...this.options.lavalink
-	});
+	public lavalink: Lavalink = DEV
```
kyranet:
```suggestion
	public lavalink: Lavalink | null = DEV
```

kyranet (changes requested):
This should use another constant (`DEV_LL`/`DEV_LAVALINK`), sometimes I need to test changes in voice, and I can't rely on production for this task.

Also, MusicCommand should be modified to disable themselves in init when `this.client.lavalink` is `null`.

config.ts.example
```diff
@@ -4,7 +4,8 @@ import { KlasaClientOptions } from 'klasa';
 import { RPoolConnectionOptions } from 'rethinkdb-ts';
 import { APIWebhookData } from './src/lib/types/DiscordAPI';
 
-export const DEV = 'DEV' in process.env ? process.env.DEV === 'true' : !('PM2_HOME' in process.env);
+const DEV = 'DEV' in process.env ? process.env.DEV === 'true' : !('PM2_HOME' in process.env);
+export const DEV_LAVALINK = true
```
kyranet:
```suggestion
export const DEV_LAVALINK = 'DEV_LAVALINK' in process.env ? process.env.DEV_LAVALINK === 'true' : DEV;
```

kyranet (changes requested):

kyranet (approved):

PR #169, opened 2019-05-30
Database init fix

Skillz4Killz:
This PR moves databaseInit.run into `klasaReady` as it requires SG to be ready before running the database init.

**Note:** It also adds a `yarn develop` script to package.json that will run the `build && start` scripts to make it easier to develop on Skyra

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,8 @@
     "test:lint": "eslint --ext ts src",
     "update": "yarn",
     "build": "tsc -p .",
-    "watch": "tsc -p . -w"
+    "watch": "tsc -p . -w",
+    "develop": "yarn build && yarn start"
   },
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "update": "yarn",
     "build": "tsc -p .",
     "watch": "tsc -p . -w",
-    "develop": "yarn build && yarn start"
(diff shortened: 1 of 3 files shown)


src/events/klasaReady.ts
```diff
@@ -34,6 +36,7 @@ export default class extends Event {
 			this.client.giveaways.init().catch(error => this.client.emit(Events.Wtf, error));
 			this.initCleanupTask().catch(error => this.client.emit(Events.Wtf, error));
 			this.initPostStatsTask().catch(error => this.client.emit(Events.Wtf, error));
+			await DatabaseInit.run(r);
```
kyranet: This should be run before everything else, specially before `this.client.giveaways.init()`, since it relies on DatabaseInit having ran.

package.json
```diff
@@ -16,7 +16,8 @@
     "test:lint": "eslint --ext ts src",
     "update": "yarn",
     "build": "tsc -p .",
-    "watch": "tsc -p . -w"
+    "watch": "tsc -p . -w",
+    "develop": "yarn build && yarn start"
```
kyranet: `dev` is a better name:
```suggestion
    "dev": "yarn build && yarn start"
```

kyranet (changes requested):

kyranet (approved):

PR #171, opened 2019-05-31
Unique role sets

Skillz4Killz:
This PR adds in the requested Unique Role Set feature. A unique role set is a set of roles where the user can only have one of them.

- [x] Adds `guild.settings.roles.uniqueRoleSets`
- [x] Adds a `s!roleset` command
- [x] Handles role removals in a raw memberUpdate event
- [x] Handles removing roles with roles command
- [x] Handles removing roles with subscribe command
- [x] Handles remvoing roles with role reaction add
- [x] All strings added to languages
- [x] Roleset Command tested!
- [x] Roles command tested
- [x] Subscribe command tested
- [x] Role reaction tested
- [x] Raw event tested


kyranet (approved):
It works! That's one hell of a job, thanks you!

PR #182, opened 2019-06-09
Fix Cat Command

Skillz4Killz:
This PR adds a new API to fix the broken cat command currently.

**Note:** I have left the old cat stuff commented out incase the old way ever works in future. Let me know if that is something you want to keep or remove.

image


kyranet (approved):
Fancy, thanks!

PR #253, opened 2019-08-26
hg: Add `--autofill` and `--autoskip` flags

gc:
Adds:
`--autofill`
Which: Adds everyone in the last 100 messages to the hunger games. Note: You can still add people alongside it, e.g. `s!hg kyra, vlad --autofill`

`--autoskip`
Which: Automatically skips through the pages so you dont have to click on the reactions, but still allows you to click on the reactions to stop, or to skip. It waits for `msg.content.length / 20` seconds, then skips.


kyranet (approved):
LGTM, thanks!

PR #258, opened 2019-08-26
Attempt to fix lgtm errors

PyroTechniac:


src/commands/Tools/quote.ts
```diff
@@ -46,7 +46,7 @@ export default class extends SkyraCommand {
 		try {
 			const parsed = new URL(url);
 			// Only discordapp.com urls are allowed.
-			if (!parsed.host.endsWith('discordapp.com')) return null;
+			if (parsed.hostname !== 'discordapp.com') return null;
```
kyranet: I meant to use RegExp instead: https://regex101.com/r/erZ3NS/1

src/lib/util/util.ts
```diff
@@ -454,7 +454,7 @@ export async function createMuteRole(message: Message) {
 }
 
 export function inlineCodeblock(input: string) {
-	return `\`${input.replace(/ /g, '\u00A0').replace(/`/g, '\\`')}\``;
+	return `\`${input.replace(/ /gs, '\u00A0').replace(/`/g, '\\`')}\``;
```
kyranet: Just curiosity, what's the effect of using the sticky flag here and how does that fix the warning?

src/lib/util/util.ts
```diff
@@ -454,7 +454,7 @@ export async function createMuteRole(message: Message) {
 }
 
 export function inlineCodeblock(input: string) {
-	return `\`${input.replace(/ /g, '\u00A0').replace(/`/g, '\\`')}\``;
+	return `\`${input.replace(/ /g, '\u00A0').replace(/`/g, '\u200B')}\``;
```
kyranet: You ate the `. :P 
```suggestion
	return `\`${input.replace(/ /g, '\u00A0').replace(/`/g, '`\u200B')}\``;
```

kyranet (changes requested):

kyranet (approved):
LGTM to me and to the LGTM bot!

PR #356, opened 2019-10-13
feat: implement RichDisplay in urban command

favna:
This way it becomes far easier to navigate through pages and since it's all just 1 API call there really is no reason to not give the option to go through all pages imo. I also made it so the definitions are properly sorted by the difference of thumbs up - thumbs down.

I was going to also add an actual "define" command for regular word definitions in this PR but apparently the (free) API I used to use is returning 429 even for the first request. Send them a mail (apparently their preferred method of contact) so once that's fixed expect another PR. That... was also meant to be the meat of this PR so this PR as it is right now is gonna look a bit barebones ¯\\\_(ツ)_/¯.


src/commands/Tools/Dictionary/urban.ts
```diff
@@ -16,43 +17,54 @@ export default class extends SkyraCommand {
...
 
 	public content(definition: string, permalink: string, i18n: Language) {
 		if (definition.length < 750) return definition;
 		return i18n.tget('SYSTEM_TEXT_TRUNCATED', cutText(definition, 750), permalink);
 	}
 
+	private buildDisplay(results: UrbanDictionaryResultOkEntry[], message: KlasaMessage, query: string) {
+		const display = new UserRichDisplay();
+
+		results.forEach(result => {
```
kyranet:
```suggestion
		for (const result of results)
```

src/commands/Tools/Dictionary/urban.ts
```diff
@@ -16,43 +17,54 @@ export default class extends SkyraCommand {
...
+					.setColor(getColor(message) || 0xFFAB2D)
+					.setThumbnail('https://i.imgur.com/CcIZZsa.png')
+					.setDescription(definition)
+					.addField('Example', example)
+					.addField('Author', result.author)
+					.addField(ZWS, `\\👍 ${result.thumbs_up}`, true)
+					.addField(ZWS, `\\👎 ${result.thumbs_down}`, true)
+					.setFooter('© Urban Dictionary')
+			);
+		});
```
kyranet:
```suggestion
		}
```

kyranet (changes requested):

src/commands/Tools/Dictionary/urban.ts
```diff
@@ -16,43 +17,54 @@ export default class extends SkyraCommand {
...
 	}
 
-	public async run(message: KlasaMessage, [query, ind = 1]: [string, number]) {
-		const index = ind - 1;
-		if (index < 0) {
-			throw message.language.tget('RESOLVER_POSITIVE_AMOUNT');
-		}
+	public async run(message: KlasaMessage, [query]: [string]) {
+		const response = await message.sendEmbed(
+			new MessageEmbed({ description: message.language.tget('SYSTEM_LOADING'), color: getColor(message) || 0xFFAB2D })
```
kyranet: Always prefer constructor arguments:

```suggestion
			new MessageEmbed()
				.setDescription(message.language.tget('SYSTEM_LOADING'))
				.setColor(getColor(message) || 0xFFAB2D)
```

kyranet (changes requested):

src/commands/Tools/Dictionary/urban.ts
```diff
@@ -16,43 +17,54 @@ export default class extends SkyraCommand {
...
-		const { list } = await fetch(`https://api.urbandictionary.com/v0/define?term=${encodeURIComponent(query)}`, 'json') as UrbanDictionaryResultOk;
+		let { list } = await fetch(`https://api.urbandictionary.com/v0/define?term=${encodeURIComponent(query)}`, 'json') as UrbanDictionaryResultOk;
 
-		const result = list[index];
-		if (typeof result === 'undefined') {
-			throw index === 0
-				? message.language.tget('COMMAND_URBAN_NOTFOUND')
-				: message.language.tget('COMMAND_URBAN_INDEX_NOTFOUND');
-		}
+		list = list.sort((a, b) => b.thumbs_up - b.thumbs_down - (a.thumbs_up - a.thumbs_down));
```
kyranet: Instead of mutating the variable, I would prefer the first variable to be a constant called `result`, and then do

```typescript
const list = result.list.sort((a, b) => b.thumbs_up - b.thumbs_down - (a.thumbs_up - a.thumbs_down));
```

kyranet (changes requested):

kyranet (approved):

PR #366, opened 2019-10-18
feat(anime): implement UserDisplay for Anime/Manga

favna:
This is the first PR of many to come, as you requested they'll be in smaller batches. Bundled these 2 commands since the files were and still are so similar anyway. Please do verify the Spanish translations I added, the only Spanish I know is `ne hablo espanol` and I don't fully trust Google Translate 😂.

I marked `Attributes.startDate` and `Attributes.endDate` as optional and handled the former in the translation as I noticed that they can be undefined at least on manga's, which caused this:
00-40-16_19-10-2019_DiscordCanary

Edit: Rewritten to use Algolia search API which is Kitsu's preferred API

### DIFF
--- a/src/commands/Anime/manga.ts
+++ b/src/commands/Anime/manga.ts
@@ -24,10 +24,7 @@ export default class extends SkyraCommand {
 			.setDescription(message.language.tget('SYSTEM_LOADING'))
 			.setColor(getColor(message) || 0xFFAB2D));
 
-		const url = new URL(`https://${TOKENS.KITSU.ID}-dsn.algolia.net/1/indexes/production_media/query`);
-
-
-		const { hits: entries } = await fetch(url, {
+		const { hits: entries } = await fetch(`https://${TOKENS.KITSU.ID}-dsn.algolia.net/1/indexes/production_media/query`, {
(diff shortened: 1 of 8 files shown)


src/commands/Anime/manga.ts
```diff
@@ -72,6 +51,34 @@ export default class extends SkyraCommand {
 		return total ? ((total / (max * 20)) * 100).toFixed(2) : '--.--';
 	}
 
+	private buildDisplay(entries: (Kitsu.Datum<Kitsu.MangaAttributes>)[], message: KlasaMessage) {
+		const display = new UserRichDisplay();
+
+		for (const entry of entries) {
+			this.client.console.log(entry.attributes);
+			this.client.console.error('===========');
```
kyranet: 👀

favna: woops LOL

favna: it'll be gone after the algolia rewrite dw

kyranet (commented):
LGTM, I **love** the new enhancements!

kyranet (approved):

PR #370, opened 2019-10-20
feat: add eshop command

favna:
Added Nintendo eShop command. This can find games / apps for Switch, WiiU and 3DS. Depends on the config changes I PR'd yesterday in #367.

Other notable changes:
- Renamed `COMMAND_ANIME_QUERY_FAILED` to `SYSTEM_QUERY_FAILED` so it can be referenced in other commands that make REST calls and not be super confusing or duplicated.
- pulled in @favware/unescape to avoid having to maintain a huge map of HTML entities inside Skyra (each entry comes in 2-3 combinations due to having an HTML name, Hex representation and Decimal representation). Rather than maintaining this huge list inside Skyra and considering I've already been doing so inside my own package for a long long time now I figured it would honestly be easier to just pull in my package. ~~so basically please don't shoot me kthx~~ Supported characters here: https://github.com/favware/node-packages/blob/master/packages/unescape/src/charactermap.ts
  - Charactermap will increase after additions that will also change this from WIP to definite.
  - Also it doesn't all that often get updates so you won't get dependabot spam for this package. Since itself has no dependencies I only really need to update it when I add new entities / get some request for it.


src/commands/Tools/eshop.ts
```diff
@@ -0,0 +1,144 @@
...
+					.setThumbnail(`https://nintendo.com${game.boxArt}`)
+					.setDescription(description)
+					.addField('Price', price, true)
+					.addField('Availability', game.availability[0], true)
+					.addField('Release Date', game.releaseDateMask === 'TBD' ? game.releaseDateMask : new Timestamp('MMMM d YYYY').displayUTC(game.releaseDateMask), true)
+					.addField('Number of Players', util.toTitleCase(game.players), true)
+					.addField('Platform', game.platform, true)
+					.addField('NSUID', game.nsuid ? game.nsuid : 'TBD', true)
+					.addField('ESRB', esrbText)
+					.addField('Categories', game.categories.join(', '))
```
favna: Not sure how to best handle translations for the field titles

src/commands/Tools/eshop.ts
```diff
@@ -0,0 +1,144 @@
...
+		const display = new UserRichDisplay();
+
+		for (const game of entries) {
+			const description = decode(cutText(game.description.replace(/\s\n {2,}/g, ' '), 750));
+			const esrbText = game.esrb
+			 ? [
+				 `**${game.esrb}**`,
+				  game.esrbDescriptors && game.esrbDescriptors.length ? ` - ${game.esrbDescriptors.join(', ')}` : ''
+				].join('')
+			 : 'Not in database';
```
favna: Not sure how to best handle translation here

favna (commented):
Just 2 concerns I still have

src/commands/Tools/eshop.ts
kyranet: Like this:

https://github.com/kyranet/Skyra/blob/2c800f6d93b28b277257bb6d5542938b5b2db42f/src/languages/en-US.ts#L2115-L2119

src/commands/Tools/eshop.ts
kyranet: Like this:

```suggestion
			 : message.language.tget('COMMAND_ESHOP_NOT_IN_DATABASE');
```

src/lib/types/Languages.d.ts
```diff
@@ -501,6 +501,10 @@ export interface LanguageKeys {
 	COMMAND_CONTENT_EXTENDED: string;
 	COMMAND_EMOJI_DESCRIPTION: string;
 	COMMAND_EMOJI_EXTENDED: string;
+	COMMAND_ESHOP_DESCRIPTION: string;
+	COMMAND_ESHOP_EXTENDED: string;
+	COMMAND_ESHOP_NOT_IN_DATABASE: string;
+	COMMAND_ESHOP_TITLES: Record<string, string>;
```
kyranet: Please strict type this here, not in the command.

kyranet (changes requested):

src/commands/Tools/eshop.ts
```diff
@@ -0,0 +1,153 @@
...
+			display.addPage(
+				new MessageEmbed()
+					.setColor(getColor(message) || 0xFFA600)
+					.setTitle(game.title)
+					.setURL(`https://nintendo.com${game.url}`)
+					.setThumbnail(`https://nintendo.com${game.boxArt}`)
+					.setDescription(description)
+					.addField(titles.PRICE, price, true)
+					.addField(titles.AVAILABILITY, game.availability[0], true)
+					.addField(titles.RELEASE_DATE, game.releaseDateMask === 'TBD' ? game.releaseDateMask : new Timestamp('MMMM d YYYY').displayUTC(game.releaseDateMask), true)
```
kyranet: Please cache this Timestamp instance.

src/commands/Tools/eshop.ts
```diff
@@ -0,0 +1,153 @@
...
+					.addField(titles.NUMBER_OF_PLAYERS, util.toTitleCase(game.players), true)
+					.addField(titles.PLATFORM, game.platform, true)
+					.addField('NSUID', game.nsuid ? game.nsuid : 'TBD', true)
```
kyranet:
```suggestion
					.addField('NSUID', game.nsuid || 'TBD', true)
```

kyranet: And also move NSUID to i18n.

src/commands/Tools/eshop.ts
```diff
@@ -0,0 +1,153 @@
...
+					.addField('ESRB', esrbText)
```
kyranet: Move ESRB to i18n.

src/commands/Tools/eshop.ts
```diff
@@ -0,0 +1,153 @@
...
+			const titles = message.language.language.COMMAND_ESHOP_TITLES as unknown as EshopLanguage;
+			const description = decode(cutText(game.description.replace(/\s\n {2,}/g, ' '), 750));
+			const esrbText = game.esrb
+				? [
+					`**${game.esrb}**`,
+					game.esrbDescriptors && game.esrbDescriptors.length ? ` - ${game.esrbDescriptors.join(', ')}` : ''
+				].join('')
+				: message.language.tget('COMMAND_ESHOP_NOT_IN_DATABASE');
+			let price = 'Free';
+			if (game.msrp && game.msrp > 0) price = `$${game.msrp} USD`;
```
kyranet:
```suggestion
			const price = game.msrp > 0 ? `$${game.msrp} USD` : 'Free';
```

But also move this to i18n.

src/commands/Tools/eshop.ts
```diff
@@ -0,0 +1,153 @@
...
+
+			display.addPage(
+				new MessageEmbed()
+					.setColor(getColor(message) || 0xFFA600)
```
kyranet: This line can be moved to line 63 to avoid re-getting the color.

kyranet (changes requested):

kyranet (approved):

PR #384, opened 2019-10-22
Fix Daily bug

PyroTechniac:
Potential fix for the daily command bug but I am unable to test it as of now.


kyranet:
This is a bad workaround though, the problem could be in the provider and this would ignore the real issue.

PyroTechniac:
I'll look into it further when I'm able to test the issue

kyranet:
Resolved in 5d5e069.

PR #389, opened 2019-10-22
Add a hint for the help command's --all flag

### LINKED ISSUE #322: Help: Add `Use the --all flag for the list in DMs` together with the embed in the help command.

tech6hutch:
It displays the message in the footer of the help embed.

E.g., "2/2 | Use the --all flag to get the full list in DMs":
image

Tested locally, and works.

Resolves #322

### DIFF
--- a/src/commands/General/Chat Bot Info/help.ts
+++ b/src/commands/General/Chat Bot Info/help.ts
@@ -73,7 +73,7 @@ export default class extends SkyraCommand {
 		const prefix = message.guildSettings.get(GuildSettings.Prefix);
 
 		const display = new UserRichDisplay()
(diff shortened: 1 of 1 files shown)


kyranet (approved):
The idea was to make it in the content, but since [Jake is thinking of enabling (at least a subset)](https://github.com/discordapp/discord-api-docs/issues/1087#issuecomment-524656071) of embeds sent by bots, I guess it's fine.

tech6hutch:
I was at first considering putting it in the content, but I wasn't sure where that would be done, since the initial content is the loading text. I hope that "show embed" idea is implemented.

PR #398, opened 2019-10-22
Feat: ApplyOptions and CreateResolver decorators.

PyroTechniac:


src/lib/util/util.ts
```diff
@@ -547,6 +547,30 @@ export function enumerable(value: boolean) {
...
+	return (target: Constructor<Command>) => class extends target {
+
+		public constructor(store: CommandStore, file: string[], directory: string, options: CommandOptions) {
+			super(store, file, directory, options);
+
+			this.createCustomResolver(name, fn);
+		}
+
+	};
+}
```
kyranet, line 572: Those should be moved to https://github.com/kyranet/klasa-decorators
And as a matter of a fact, @Favna is already working on it: https://github.com/kyranet/klasa-decorators/pull/19

kyranet: Also, I'm not very fond of having to... creating a class every time those are used, we should investigate a way to do something like `@Batch` which implements an array as a non-enumerable static symbol-keyed property that is called when the constructor is, likewise how Klasa's plugin system works.

That needs some investigation, I know it's completely possible and to be honest, it should be done over having to create multiple classes dynamically, which take a lot of memory and makes inheritance checks significantly slower.

PyroTechniac:
Closing until Favna's and my PRs are merged in klasa-decorators

PR #404, opened 2019-10-23
Language fix

PyroTechniac:


PyroTechniac:
Didn't know if I should go with username or mention, so I went with username for now

kyranet (approved):

PR #429, opened 2019-10-26
feat: tmdb commands

favna:
This adds 2 new commands - movies and shows. The first is for searching
movies on TheMovieDatabase and the latter for TV shows.

***config change required***

This requires an API key for TheMovieDatabase.

1. Go to https://themoviedb.org
2. Create account
3. Go to https://themoviedb.org/settings/api
4. Create app for API key
5. Copy the API Key (v3 auth) (v4 only does user data)


favna:
Gave it SEM: Minor because it requires a config change thus a reboot. Not sure when you do those exactly.

kyranet (approved):

PR #431, opened 2019-10-27
feat: reddit user command

favna:
Gets stats for Reddit users. Next one in the list of commands I had to port. 

image

image

### DIFF
--- a/src/commands/Misc/reddituser.ts
+++ b/src/commands/Misc/reddituser.ts
@@ -1,0 +1,238 @@
+import { Collection, MessageEmbed } from 'discord.js';
+import { CommandStore, KlasaMessage, Timestamp } from 'klasa';
+import { SkyraCommand } from '../../lib/structures/SkyraCommand';
+import { Reddit } from '../../lib/types/definitions/Reddit';
+import { BrandingColors } from '../../lib/util/constants';
+import { cutText, fetch, getColor, roundNumber } from '../../lib/util/util';
+
+const COMPLEXITY_LEVELS = ['very low', 'low', 'medium', 'high', 'very high', 'very high'];
+
+export default class extends SkyraCommand {
+
+	private joinedRedditTimestamp = new Timestamp('MMMM d YYYY');
+	private usernameRegex = /^(?:\/?u\/)?[A-Za-z0-9_-]*$/;
+
+	public constructor(store: CommandStore, file: string[], directory: string) {
+		super(store, file, directory, {
+			aliases: ['redditor'],
+			cooldown: 10,
+			description: language => language.tget('COMMAND_REDDITUSER_DESCRIPTION'),
+			extendedHelp: language => language.tget('COMMAND_REDDITUSER_EXTENDED'),
+			requiredPermissions: ['EMBED_LINKS'],
+			usage: '<user:user>'
+		});
+
+		this.createCustomResolver('user', (arg, _possible, message) => {
+			if (!this.usernameRegex.test(arg)) throw message.language.tget('COMMAND_REDDITUSER_INVALID_USER', arg);
+			arg = arg.replace(/^\/?u\//, '');
+			return arg;
+		});
+	}
+
+	public async run(message: KlasaMessage, [user]: [string]) {
+		await message.sendEmbed(new MessageEmbed()
+			.setDescription(message.language.tget('SYSTEM_LOADING'))
+			.setColor(BrandingColors.Secondary));
+
+		const [about, comments, posts] = await this.fetchData(user, message);
+		if (!about || !comments || !posts || !comments.length || !posts.length) throw message.language.tget('COMMAND_REDDITUSER_QUERY_FAILED');
+		comments.sort((a, b) => b.score - a.score);
+
+		const titles = message.language.tget('COMMAND_REDDITUSER_TITLES');
+		const fieldsData = message.language.tget('COMMAND_REDDITUSER_DATA');
+		const [bestComment] = comments;
+		const worstComment = comments[comments.length - 1];
+		const complexity = roundNumber(this.calculateTextComplexity(comments), 2);
+
+		const embed = new MessageEmbed()
+			.setTitle(fieldsData.OVERVIEW_FOR(about.name))
+			.setURL(`https://www.reddit.com${about.subreddit.url}`)
+			.setColor(getColor(message))
+			.setDescription(fieldsData.JOINED_REDDIT(this.joinedRedditTimestamp.displayUTC(about.created)))
+			.setThumbnail(about.icon_img)
+			.setFooter(fieldsData.DATA_AVAILABLE_FOR)
+			.addField(titles.LINK_KARMA, about.link_karma, true)
+			.addField(titles.COMMENT_KARMA, about.comment_karma, true)
+			.addField(titles.TOTAL_COMMENTS, comments.length, true)
+			.addField(titles.TOTAL_SUBMISSIONS, posts.length, true)
+			.addField(titles.COMMENT_CONTROVERSIALITY, `${roundNumber(this.calculateControversiality(comments), 1)}%`, true)
+			.addField(titles.TEXT_COMPLEXITY, `${COMPLEXITY_LEVELS[Math.floor(complexity / 20)]} (${roundNumber(complexity, 1)}%)`, true)
+			.addField(`${titles.TOP_5_SUBREDDITS} (${titles.BY_SUBMISSIONS})`, this.calculateTopContribution(posts), true)
+			.addField(`${titles.TOP_5_SUBREDDITS} (${titles.BY_COMMENTS})`, this.calculateTopContribution(comments), true)
+			.addField(titles.BEST_COMMENT,
+				[
+					`${bestComment.subreddit} - **${bestComment.score}**`,
+					`${message.language.duration((Date.now() - bestComment.created) * 1000)} ago`,
+					`[${fieldsData.PERMALINK}](https://reddit.com${bestComment.permalink})`,
+					cutText(bestComment.body, 900)
+				].join('\n'))
+			.addField(titles.WORST_COMMENT,
+				[
+					`${worstComment.subreddit} - **${worstComment.score}**`,
+					`${message.language.duration((Date.now() - worstComment.created) * 1000)} ago`,
+					`[${fieldsData.PERMALINK}](https://reddit.com${worstComment.permalink})`,
+					cutText(worstComment.body, 900)
+				].join('\n'));
+
+		return message.sendEmbed(embed);
+	}
+
+	private async fetchData(user: string, message: KlasaMessage) {
+		return Promise.all([this.fetchAbout(user, message), this.fetchComments(user, message), this.fetchPosts(user, message)]);
+	}
+
+	private async fetchAbout(user: string, message: KlasaMessage) {
+		const { data } = await fetch(`https://www.reddit.com/user/${user}/about/.json`, 'json')
+			.catch(() => {
+				throw message.language.tget('COMMAND_REDDITUSER_QUERY_FAILED');
+			}) as Reddit.Response<'about'>;
+		return data;
+	}
+
+	/**
+	 * Fetches comments from the Reddit API
+	 * This is a recursive function as Reddit API can only give comments in sets of 100 with a maximum of 1000
+	 * @param user The reddit user to get comments for
+	 * @param message The discord message that triggered this command
+	 * @param after Recursive parameter, determines after which comment to start fetching
+	 * @param dataElements Recursive parameter, the previously fetched comments to retain a total
+	 */
+	private async fetchComments(user: string, message: KlasaMessage, after = '', dataElements: Reddit.CommentDataElement[] = []) {
+		const url = new URL(`https://www.reddit.com/user/${user}/comments.json`);
+		url.searchParams.append('after', after);
+		url.searchParams.append('limit', '100');
+
+		const { data } = await fetch(url, 'json')
+			.catch(() => {
+				throw message.language.tget('COMMAND_REDDITUSER_QUERY_FAILED');
+			}) as Reddit.Response<'comments'>;
+
+		for (const child of data.children) {
+			dataElements.push(child.data);
+		}
+
+		if (data.children.length === 100) await this.fetchComments(user, message, data.children[99].data.name, dataElements);
+
+		return dataElements;
+	}
+
+	/**
+	 * Fetches posts from the Reddit API
+	 * This is a recursive function as Reddit API can only give posts in sets of 100
+	 * @param user The reddit user to get posts for
+	 * @param message The discord message that triggered this command
+	 * @param after Recursive parameter, determines after which post to start fetching
+	 * @param dataElements Recursive parameter, the previously fetched posts to retain a total
+	 */
+	private async fetchPosts(user: string, message: KlasaMessage, after = '', dataElements: Reddit.PostDataElement[] = []) {
+		const url = new URL(`https://www.reddit.com/user/${user}/submitted.json`);
+		url.searchParams.append('after', after);
+		url.searchParams.append('limit', '100');
+
+		const { data } = await fetch(url, 'json')
+			.catch(() => {
+				throw message.language.tget('COMMAND_REDDITUSER_QUERY_FAILED');
+			}) as Reddit.Response<'posts'>;
+
+		for (const child of data.children) {
+			dataElements.push(child.data);
+		}
+
+		if (data.children.length === 100) await this.fetchPosts(user, message, data.children[99].data.name, dataElements);
+
+		return dataElements;
+	}
+
+	/**
+	 * Calculated the controversiality based on whether the Reddit has marked a comment controversial
+	 * Controversial posts or comments have very even amounts of up/downvotes and a lot of activity.
+	 * @param comments The reddit comments to take into the calculation
+	 */
+	private calculateControversiality(comments: Reddit.CommentDataElement[]) {
+		if (!comments.length || comments.length < 5) return 0;
+		let count = 0;
+
+		for (const comment of comments) {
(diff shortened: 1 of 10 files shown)


src/commands/Misc/reddituser.ts
```diff
@@ -0,0 +1,233 @@
...
+import { fetch, roundNumber, getColor, cutText } from '../../lib/util/util';
+import { Reddit } from '../../lib/types/definitions/Reddit';
+import { MessageEmbed } from 'discord.js';
+import { BrandingColors } from '../../lib/util/constants';
+
+const COMPLEXITY_LEVELS = ['very low', 'low', 'medium', 'high', 'very high', 'very high'];
+
+export default class extends SkyraCommand {
+
+	private comments: Reddit.CommentDataElement[];
```
vladfrangu: You should really not set this as a class property... If two users run it at the same time, you can get information clashing...

src/commands/Misc/reddituser.ts
```diff
@@ -0,0 +1,233 @@
...
+			.setDescription(message.language.tget('SYSTEM_LOADING'))
+			.setColor(BrandingColors.Secondary));
+
+		await this.fetchData(user);
+		if (!this.comments.length || !this.submitted.length) throw message.language.tget('COMMAND_REDDITUSER_QUERY_FAILED');
+		this.comments.sort((a, b) => b.score - a.score);
+
+		const titles = message.language.tget('COMMAND_REDDITUSER_TITLES');
+		const fieldsData = message.language.tget('COMMAND_REDDITUSER_DATA');
+		const bestComment = this.comments[0];
```
vladfrangu:
```suggestion
		const [bestComment] = this.comments;
```

src/commands/Misc/reddituser.ts
```diff
@@ -0,0 +1,233 @@
...
+
+		return syl ? syl.length : 1;
+	}
+
+	private calculateTextComplexity() {
+		let sentenceCount = 0;
+		let syllableCount = 0;
+		let wordCount = 0;
+
+		this.comments.forEach(item => {
```
vladfrangu: This can be turned into a for-of loop

src/commands/Misc/reddituser.ts
```diff
@@ -0,0 +1,233 @@
...
+			const sentences = item.body.split(/[.!?]+/gm);
+			const words = item.body.trim().split(/\s+/gm);
+
+			sentenceCount += sentences.length - 1;
+			wordCount += words.length;
+
+			words.forEach(wordSyl => {
```
vladfrangu: This can be a for-of loop

src/commands/Misc/reddituser.ts
```diff
@@ -0,0 +1,233 @@
...
+		});
+
+		return this.calculateKincaid(sentenceCount, wordCount, syllableCount);
+	}
+
+	private calculateTopSubredditsSubmissions() {
+		const subredditCounts = [];
+		const subreddits: { [subRedditName: string]: number } = {};
+
+		this.submitted.forEach(item => {
```
vladfrangu:
```ts
for (const item of submitted) {
	subreddits[item.subreddit] = (subreddits[item.subreddit] || 0) + 1;
}
```

src/commands/Misc/reddituser.ts
```diff
@@ -0,0 +1,233 @@
...
+				syllableCount += this.calculateSyllables(wordSyl.toLowerCase());
+			});
+		});
+
+		return this.calculateKincaid(sentenceCount, wordCount, syllableCount);
+	}
+
+	private calculateTopSubredditsSubmissions() {
+		const subredditCounts = [];
+		const subreddits: { [subRedditName: string]: number } = {};
```
vladfrangu:
```suggestion
		const subreddits: Record<string, number> = {};
```

src/commands/Misc/reddituser.ts
```diff
@@ -0,0 +1,233 @@
...
+		this.submitted.forEach(item => {
+			if (subreddits.hasOwnProperty(item.subreddit)) {
+				subreddits[item.subreddit] += 1;
+			} else {
+				subreddits[item.subreddit] = 1;
+			}
+		});
+
+		// eslint-disable-next-line guard-for-in
+		for (const subreddit in subreddits) {
```
vladfrangu:
```suggestion
		for (const [subreddit, count] of Object.entries(subreddits)) {
```

src/commands/Misc/reddituser.ts
```diff
@@ -0,0 +1,233 @@
...
+			subredditCounts.push({
+				count: subreddits[subreddit],
+				name: subreddit
+			});
+		}
+
+		subredditCounts.sort((a, b) => b.count - a.count);
+		subredditCounts.splice(5);
```
vladfrangu: Whats the point of this? Removing the first 5 elements cause...?

src/commands/Misc/reddituser.ts
```diff
@@ -0,0 +1,233 @@
...
+			(val, index) => `**${index + 1}:** [/r/${val.name}](https://wwww.reddit.com/r/${val.name}) (${val.count})`
+		)
+			.join('\n');
+	}
+
+	private calculateTopSubredditsComments() {
+		const subredditCounts = [];
+		const subreddits: { [subRedditName: string]: number } = {};
+
+		this.comments.forEach(item => {
```
vladfrangu: As above

src/lib/util/util.ts
```diff
@@ -575,7 +590,7 @@ export function ratelimit(bucket: number, cooldown: number, auth = false) {
 
 			try {
 				bucket.drip();
-			} catch {}
+			} catch { }
```
vladfrangu, line 439: Why? Linter error or?

vladfrangu (changes requested):

src/lib/util/util.ts
favna: Might've been VSCode's autoformatter as well. I didn't actively change it.

src/commands/Misc/reddituser.ts
```diff
@@ -0,0 +1,212 @@
...
+import { fetch, roundNumber, getColor, cutText } from '../../lib/util/util';
+import { Reddit } from '../../lib/types/definitions/Reddit';
+import { MessageEmbed } from 'discord.js';
+import { BrandingColors } from '../../lib/util/constants';
+
+const COMPLEXITY_LEVELS = ['very low', 'low', 'medium', 'high', 'very high', 'very high'];
+
+export default class extends SkyraCommand {
+
+	private joinedRedditTimestamp = new Timestamp('MMMM d YYYY');
```
vladfrangu, line 10: Is there any reason for this to be a class property?

src/commands/Misc/reddituser.ts
```diff
@@ -0,0 +1,212 @@
...
+			description: language => language.tget('COMMAND_REDDITUSER_DESCRIPTION'),
+			extendedHelp: language => language.tget('COMMAND_REDDITUSER_EXTENDED'),
+			requiredPermissions: ['EMBED_LINKS'],
+			usage: '<user:str>'
+		});
+	}
+
+	public async run(message: KlasaMessage, [user]: [string]) {
+		user = user.replace(/^\/?u\//, '');
+		if (!/[A-z0-9_-]/.test(user)) throw message.language.tget('COMMAND_REDDITUSER_INVALID_USER', user);
```
vladfrangu: Nit: you can move this into a custom resolver

vladfrangu (changes requested):

src/commands/Misc/reddituser.ts
kyranet: Consistency with other pieces.

src/commands/Misc/reddituser.ts
```diff
@@ -0,0 +1,219 @@
...
+		return dataElements;
+	}
+
+	private async fetchSubmissions(user: string, message: KlasaMessage, after = '', dataElements: Reddit.PostDataElement[] = []) {
+		const url = new URL(`https://www.reddit.com/user/${user}/submitted.json`);
+		url.searchParams.append('after', after);
+		url.searchParams.append('limit', '100');
+
+		const { data } = await fetch(url, 'json')
+		.catch(() => { throw message.language.tget('COMMAND_REDDITUSER_QUERY_FAILED'); }) as Reddit.Response<'posts'>;
```
kyranet:
```suggestion
			.catch(() => { throw message.language.tget('COMMAND_REDDITUSER_QUERY_FAILED'); }) as Reddit.Response<'posts'>;
```

kyranet (changes requested):

src/commands/Misc/reddituser.ts
```diff
@@ -0,0 +1,219 @@
...
+import { Reddit } from '../../lib/types/definitions/Reddit';
+import { MessageEmbed } from 'discord.js';
+import { BrandingColors } from '../../lib/util/constants';
+
+const COMPLEXITY_LEVELS = ['very low', 'low', 'medium', 'high', 'very high', 'very high'];
+
+export default class extends SkyraCommand {
+
+	private joinedRedditTimestamp = new Timestamp('MMMM d YYYY');
+	private usernameRegex = new RegExp('^(?:\/u\/)?[A-z0-9_-]*$');
```
kyranet:
```suggestion
	private usernameRegex = /^(?:/?u/)?[A-Za-z0-9_-]*$/;
```

kyranet (changes requested):

vladfrangu (dismissed):

src/commands/Misc/reddituser.ts
```diff
@@ -0,0 +1,219 @@
+import { CommandStore, KlasaMessage, Timestamp } from 'klasa';
+import { SkyraCommand } from '../../lib/structures/SkyraCommand';
+import { fetch, roundNumber, getColor, cutText } from '../../lib/util/util';
+import { Reddit } from '../../lib/types/definitions/Reddit';
+import { MessageEmbed } from 'discord.js';
+import { BrandingColors } from '../../lib/util/constants';
+
+const COMPLEXITY_LEVELS = ['very low', 'low', 'medium', 'high', 'very high', 'very high'];
```
kyranet: This should be moved to i18n.

src/commands/Misc/reddituser.ts
```diff
@@ -0,0 +1,219 @@
...
+			.catch(() => { throw message.language.tget('COMMAND_REDDITUSER_QUERY_FAILED'); }) as Reddit.Response<'comments'>;
+
+		for (const child of data.children) {
+			dataElements.push(child.data);
+		}
+
+		if (data.children.length === 100) await this.fetchComments(user, message, data.children[99].data.name, dataElements); // Recurse to get more data
+
+		return dataElements;
+	}
```
kyranet: This method needs to be documented clarifying that it's a recursive method.

src/commands/Misc/reddituser.ts
```diff
@@ -0,0 +1,219 @@
...
+			.catch(() => { throw message.language.tget('COMMAND_REDDITUSER_QUERY_FAILED'); }) as Reddit.Response<'posts'>;
+
+		for (const child of data.children) {
+			dataElements.push(child.data);
+		}
+
+		if (data.children.length === 100) await this.fetchSubmissions(user, message, data.children[99].data.name, dataElements); // Recurse to get more data
+
+		return dataElements;
+	}
```
kyranet: This method needs to be documented clarifying that it's a recursive method.

src/commands/Misc/reddituser.ts
```diff
@@ -0,0 +1,219 @@
...
+	private calculateControversiality(comments: Reddit.CommentDataElement[]) {
+		if (!comments.length || comments.length < 5) return 0;
+		let count = 0;
+
+		for (const comment of comments) {
+			if (comment.controversiality === 1) count++;
+		}
+
+		return (count / comments.length) * 100;
+	}
```
kyranet, line 159: This method needs to be documented clarifying what exactly the algorithm does and what each condition means.

src/commands/Misc/reddituser.ts
```diff
@@ -0,0 +1,219 @@
...
+
+	private calculateKincaid(sentences: number, words: number, syllables: number) {
+		const sentenceWeight = 0.39;
+		const wordWeight = 11.8;
+		const adjustment = 15.59;
```
kyranet, line 171: What are those variables based on? Needs source and documentation.

src/commands/Misc/reddituser.ts
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	private calculateSyllables(word: string) {
+		word = word.toLowerCase();
+		if (word.length <= 3) return 1;
+		word = word.replace(/(?:[^laeiouy]es|ed|[^laeiouy]e)$/, '');
+		word = word.replace(/^y/, '');
+		const syl = word.match(/[aeiouy]{1,2}/g);
+
+		return syl ? syl.length : 1;
```
kyranet: `String.prototype.replace` returns the modified string, chain it:
```typescript
const syl = word.replace(/(?:[^laeiouy]es|ed|[^laeiouy]e)$/, '')
	.replace(/^y/, '')
	.match(/[aeiouy]{1,2}/g);
```

kyranet: And while this is complex, please put a note above to clarify what the algorithm actually does and some examples of why this works.

src/commands/Misc/reddituser.ts
```diff
@@ -0,0 +1,219 @@
...
+		for (const post of posts) {
+			subreddits[post.subreddit] = (subreddits[post.subreddit] || 0) + 1;
+		}
+
+		for (const [subreddit, count] of Object.entries(subreddits)) {
+			subredditCounts.push({
+				count,
+				name: subreddit
+			});
+		}
```
kyranet: This should use `Map<string, number>`, not `Record<string, number>`. Re-compiling and shape mutating are discouraged in this performance to avoid performance drawbacks.

src/commands/Misc/reddituser.ts
```diff
@@ -0,0 +1,219 @@
...
+
+		return subredditCounts.sort((a, b) => b.count - a.count).slice(0, 5).map(
+			(val, index) => `**${index + 1}:** [/r/${val.name}](https://wwww.reddit.com/r/${val.name}) (${val.count})`
+		)
+			.join('\n');
```
kyranet: Cleaner alternative with shorter lines:
```typescript
subredditCounts.sort((a, b) => b.count - a.count).slice(0, 5).map((val, index) =>
	`**${index + 1}:** [/r/${val.name}](https://wwww.reddit.com/r/${val.name}) (${val.count})`
).join('\n');
```

Though I would prefer having the map as calling a format method to reduce complexity of this method.

src/commands/Misc/reddituser.ts
```diff
@@ -0,0 +1,219 @@
...
+		for (const comment of comments) {
+			subreddits[comment.subreddit] = (subreddits[comment.subreddit] || 0) + 1;
+		}
+
+		for (const [subreddit, count] of Object.entries(subreddits)) {
+			subredditCounts.push({
+				count,
+				name: subreddit
+			});
+		}
```
kyranet: Same as above, use a `Map<string, number>`. Also, is it me or is `calculateTopSubredditsComments` full duplicated code from `calculateTopSubredditsSubmissions`?

src/commands/Misc/reddituser.ts
```diff
@@ -0,0 +1,219 @@
...
+
+		subredditCounts.sort((a, b) => b.count - a.count).slice(0, 5).map(
+			(val, index) => `**${index + 1}:** [/r/${val.name}](https://wwww.reddit.com/r/${val.name}) (${val.count})`
+		)
+			.join('\n');
```
kyranet: Same as above, use a format method.

kyranet (changes requested):

src/commands/Misc/reddituser.ts
favna: I added the comment `Calculated the controversiality based on whether the Reddit has marked a comment controversial` but it should be noted that controversiality is a reddit thing and not necessarily algorithmic. This is just a basic percentile calculation and the API can have some comments marked as controversial. Reddit defines controversial as: **Controversial are posts or comments that have very even amounts of up/downvotes and a lot of activity.**

favna: Update: added this:

```
/**
 * Calculated the controversiality based on whether the Reddit has marked a comment controversial
 * Controversial posts or comments have very even amounts of up/downvotes and a lot of activity.
 * @param comments The reddit comments to take into the calculation
 */
```

src/commands/Misc/reddituser.ts
favna: https://en.wikipedia.org/wiki/Flesch–Kincaid_readability_tests#Flesch–Kincaid_grade_level + added to the file

src/commands/Misc/reddituser.ts
```diff
@@ -0,0 +1,235 @@
...
+				syllableCount += this.calculateSyllables(syllable.toLowerCase());
+			}
+		}
+
+		return this.calculateKincaid(sentenceCount, wordCount, syllableCount);
+	}
+
+	private calculateTopContribution(contributions: Reddit.PostDataElement[] | Reddit.CommentDataElement[]) {
+		const subredditCounts = [];
+		const subreddits: Map<string, number> = new Map();
```
kyranet:
```suggestion
		const subreddits = new Map<string, number>();
```

kyranet (changes requested):

favna:
471adc4 sets the TSC version to workspace version for VSCode. Not sure how that work on Linux though because the path has `\\` while linux is gonna want `/`... But AFAIK no one of the Skyra devs uses Linux / MacOS and I assume VSCode just falls back to it's own version.

kyranet (approved):

PR #453, opened 2019-10-31
Fix(Settings): Fix settings out of sync in setPrefix command.

PyroTechniac:


kyranet:
This is an invalid patch as it's something that should never happen, will be fixed in another PR.

PR #459, opened 2019-11-01
Add Waifu cmd

### LINKED ISSUE #455: Add waifu command
Simply send: `https://www.thiswaifudoesnotexist.net/example-${Math.floor(Math.random() * 199_999)}.jpg`

They are randomly generated using deeplearning from https://www.thiswaifudoesnotexist.net/

This will also deprecate the command from Ordo.

gc:
closes #455


kyranet (changes requested):
The image should be displayed with an embed for consistency with weeb, dog/kitten, and similar commands.

Also make the 199999 number a constant to avoid "magic numbers".

PR #481, opened 2019-11-06
Fix(Anime): Fix 2 reported bugs

favna:
Can't replicate the issue with not being able to access original of `entry.posterImage` but the way I fixed it should fix it if it ever occurs again. I feel like the API just derped. 

Also added a TODO comment at the same time to change it to optional chaining when LGTM and Typescript-ESLint fully support that.


src/commands/Anime/anime.ts
```diff
@@ -70,17 +70,20 @@ export default class extends SkyraCommand {
 			const type = entry.subtype;
 			const title = entry.titles.en || entry.titles.en_jp || entry.canonicalTitle || '--';
 
+			console.log(entries[0]);
```
kyranet:
```suggestion
```

kyranet (approved):

PR #484, opened 2019-11-07
feat: Added Markov

### LINKED ISSUE #454: Add markov command
markov is a fun command lots of people like using, definitely a good addition.

Heres the implementation in Ordo (which will be deprecated upon addition of this), although it could be expanded to be more advanced definitely:

```js
const MarkovChain = require('markovchain');

const messageLimitHundreds = 10;

module.exports = class extends Command {

	constructor(...args) {
		super(...args, {
			description: 'Generate a markov chain from the text chat.',
			requiredPermissions: ['READ_MESSAGE_HISTORY'],
			cooldown: 10
		});
	}

	async run(msg) {
		let messageBank = await msg.channel.messages.fetch({ limit: 100 });
		for (let i = 1; i < messageLimitHundreds; i++) {
			messageBank = messageBank.concat(
				await msg.channel.messages.fetch({ limit: 100, before: messageBank.last().id })
			);
		}

		const quotes = new MarkovChain(messageBank.map(message => message.content).join(' '));
		const chain = quotes
			.start(this.useUpperCase)
			.end(20)
			.process();
		return msg.sendMessage(chain.substring(0, 1999));
	}

	useUpperCase(wordList) {
		const tmpList = Object.keys(wordList).filter(word => word[0] >= 'A' && word[0] <= 'Z');
		return tmpList[Math.floor(Math.random() * tmpList.length)];
	}

};
```

Notes on my implementation:
- would be much better to have a centralized function for mass-fetching lots of images, for example put it on the `Channel` class, so instead of doing that messageBank thing, you'd have `msg.channel.massFetch(1000);` or something.

- to make it much more efficient, some other improvements could be made. Not sure of a *great* way to improve it, i considered just storing an 1000-long array (FI-LO) of message contents per channel, as a cache to instantly generate markov chains without need to fetch, although that obviously has the cost of memory usage.

kyranet:
Fixes #454

~~Runs take 20-30 milliseconds, which is excessive for a command (since it's fully synchronous), 3 simultaneous runs would render a full event blockage for almost a tenth of a second, which is highly unwanted. Look into workers for this either now or in the future.~~

~~Also add caching with a WeakMap so Skyra does not have to fetch 1000 messages all the time but just download and cache it for a minute, this allows simultaneous runs to happen much faster.~~


src/commands/Fun/markov.ts
```diff
@@ -0,0 +1,69 @@
...
+			.setDescription(message.language.tget('SYSTEM_LOADING'))
+			.setColor(BrandingColors.Secondary));
+
+		// Process the chain
+		const messageBank = await this.fetchMessages(message);
+		const time = new Stopwatch();
+		const contents = messageBank.map(m => m.content).join(' ');
+		const chain = new Markov()
+			.parse(contents)
+			.start(this.useUpperCase.bind(this))
```
gc: Should be binded in the constructor

kyranet: Bound*, resolved in 34ee373, thanks for the suggestion!

PR #486, opened 2019-11-09
ci(workflow): add node_modules caching

favna:
Does exactly what the title says. Using https://github.com/marketplace/actions/cache


kyranet:
I'm not seeing a performance increase in this, tsc went from 31s to 37s and eslint went from 25s to 27s compared to the previous PR.

kyranet (approved):

PR #489, opened 2019-11-09
feat: Added message spam filter

### LINKED ISSUE #445: Feat(Moderation): Add multi-message spam filter

kyranet:
Fixes #445


PyroTechniac (dismissed):
Settings menu keys need added

src/monitors/messageFilter.ts
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	protected onAlert(message: KlasaMessage) {
+		floatPromise(this, message.alert(message.language.tget('MONITOR_MESSAGEFILTER', message.author.toString())));
+	}
+
+	protected onLogMessage(message: KlasaMessage) {
+		return new MessageEmbed()
+			.splitFields(message.content)
+			.setColor(0xEFAE45)
```
favna, line 51: Disclaimer: on mobile

 Not sure what colour this is but wouldn't it be better to move all such colours to an enum (ModerationColours)?

src/lib/schemas/Guilds.ts
```diff
@@ -72,6 +72,14 @@ export default Client.defaultGuildSchema
 			.add('hardActionDuration', 'Integer', { min: 1000, configurable: false })
 			.add('thresholdMaximum', 'Integer', { 'default': 10, 'min': 0, 'max': 60, 'configurable': false })
 			.add('thresholdDuration', 'Integer', { 'default': 60000, 'min': 0, 'max': 120000, 'configurable': false }))
+		.add('messages', capitals => capitals
```
favna, line 75: I recommend naming this differently as it doesn't leave space for other monitors triggering on messages. In the same vein as message count I know Ribbon also had "duplicate text" (based on Levenshtein distance), capitals spam, emojis spam (counts) and a registry of banned words (such as f-bombs). If such systems get added to Skyra then this keyname is going to cause conflicts.

src/monitors/messageFilter.ts
```diff
@@ -0,0 +1,81 @@
...
+		if (typeof previousValue === 'undefined') {
+			const nextValue: string[] = [];
+			this.kChannels.set(message.channel as TextChannel, nextValue);
+			return nextValue;
+		}
+
+		return previousValue;
+	}
+
+	private getCount(contents: readonly string[], value: string) {
```
favna: `ReadonlyArray<string>` is better here for the Param type IMHO.

src/commands/Management/Message Filters/messageMode.ts
```diff
@@ -0,0 +1,24 @@
+import { CommandStore } from 'klasa';
+import { SelfModerationCommand } from '../../../lib/structures/SelfModerationCommand';
+import { GuildSecurity } from '../../../lib/util/Security/GuildSecurity';
+import { GuildSettings } from '../../../lib/types/settings/GuildSettings';
+
+export default class extends SelfModerationCommand {
+
+	protected $adder: keyof GuildSecurity['adders'] = 'messages';
```
favna, line 8: Why the $ sign 🤨

favna (commented):
Got a few suggestions / questions

src/lib/schemas/Guilds.ts
kyranet: Capitals spam is already implemented (called `capitals`), emoji spam will most likely be called `emojis`, and banned words is `filter`. There's actually no conflict, even with the bunch of systems I want to implement from the issues tabs.

src/monitors/messageFilter.ts
vladfrangu: Pretty sure they're the same... Will check tho

favna: They are

src/commands/Management/Message Filters/messageMode.ts
kyranet: It's a protected abstract field from SelfModerationCommand.
And because it's a keyof, the dollar kind of tells that it's an "accessor".

src/monitors/messageFilter.ts
kyranet: `readonly T[]` is preferred over `ReadonlyArray<T>` just like we use `T[]` over `Array<T>`.

src/commands/Management/Message Filters/messageMode.ts
favna: fair enough

cfanoulis (dismissed):
I mean, it LGTM.

~~not that my review is worth anything lol~~

src/monitors/messageFilter.ts
kyranet: This is common in all existing moderation monitors, should be fixed in another pull request.

favna: Created https://github.com/kyranet/Skyra/issues/491 to cover this

favna (dismissed):
LGTM

src/monitors/messageFilter.ts
kyranet: Thanks you! 👍

favna (approved):

PR #506, opened 2019-11-12
refactor: Applied most of WebStorm's enhancements.

kyranet:


src/commands/Tools/quote.ts
```diff
@@ -43,27 +44,29 @@ export default class extends SkyraCommand {
...
+		const guild = this.client.guilds.get(_guild);
+		if (guild !== message.guild!) return null;
 
-			return await (channel as TextChannel).messages.fetch(_message);
-		} catch {
-			return null;
-		}
+		const channel = guild.channels.get(_channel);
+		if (!channel) return null;
+		if (!('messages' in channel)) throw message.language.tget('RESOLVER_INVALID_CHANNEL', 'Channel');
```
PyroTechniac:
```suggestion
		if (!(channel instanceof TextChannel)) throw message.language.tget('RESOLVER_INVALID_CHANNEL', 'Channel');
```

PyroTechniac (changes requested):

src/commands/Tools/quote.ts
vladfrangu: Doesn't account for News channels ^

PyroTechniac: Fixed to account for News Channels

kyranet: Fixed in e1b69cc

vladfrangu (approved):

PR #508, opened 2019-11-13
feat: add pokedex command

favna:
- [x] Including typings
- [x] Setting up the response type
- [x] Setting up a functions that set the proper graphql queries
- [x] Setting up first command

the returned data of the evolution chain will change a bit when the API is updated for gen8 due to the change of formatting some data, but that will be a silent change API-side, not change needed on Skyra as it already accounts for the change. Specifically that rn `raichu-alola` won't show in pikachu, other than that a sample reply:


src/commands/Pokemon/pokedex.ts
```diff
@@ -0,0 +1,176 @@
...
+				headers: {
+					'Content-Type': 'application/json'
+				},
+				body: JSON.stringify({
+					query: getPokemonDetailsByFuzzy(pokemon)
+				})
+			}, 'json') as Promise<GraphQLPokemonResponse<'getPokemonDetailsByFuzzy'>>;
+		} catch (err) {
+			throw message.language.tget('SYSTEM_QUERY_FAIL');
+		}
```
kyranet: For the `catch` to ever be triggered, you must do `return await`.

src/commands/Pokemon/pokedex.ts
```diff
@@ -0,0 +1,176 @@
...
+			case 'Purple':
+				return '#A63DE8';
+			case 'Red':
+				return '#FF3232';
+			case 'White':
+				return '#E1E1E1';
+			case 'Yellow':
+				return '#FFF359';
+			default:
+				return '#FF0000';
```
kyranet: Those should return hexadecimal numbers, which is what `MessageEmbed#setColor` actually expects. For instance, change `#323232` to `0x323232`, same thing for all others.

src/commands/Pokemon/pokedex.ts
```diff
@@ -0,0 +1,176 @@
...
+							break;
+						case 'hidden':
+							abilities += `, *${ability}*`;
+							break;
+						case 'second':
+						case 'special':
+							abilities += `, ${ability}`;
+					}
+				}
+			}
```
kyranet: For ease of maintenance, you should make abilities a `string[]`, which you later join by `', '`. This will make the code more readable and easier to maintain, plus you'll be able to convert the switch into a single if:

```typescript
for (const [type, ability] of Object.entries(poke.abilities)) {
	if (ability === null) continue;
	abilities.push(type === 'hidden' ? `*${ability}*` : ability);
}
```

src/commands/Pokemon/pokedex.ts
```diff
@@ -0,0 +1,176 @@
...
+			for (const [gender, ratio] of Object.entries(poke.gender)) {
+				switch (gender) {
+					case 'male':
+						genders += `${ratio} ♂ `;
+						break;
+					case 'female':
+						genders += ` | ${ratio} ♀`;
+						break;
+				}
+			}
```
kyranet: This looks like the case in which they can be simplified to not use a loop:
```typescript
const genders = `${poke.gender.male} ♂ | {poke.gender.female} ♀`;
```

src/commands/Pokemon/pokedex.ts
```diff
@@ -0,0 +1,176 @@
...
+			// Parse pre-evolutions and add to evochain
+			if (poke.preevolutions?.length) {
+				const { evolutionLevel } = poke.preevolutions[0];
+				evoChain = `\`${toTitleCase(poke.preevolutions[0].species)}\` ${evolutionLevel ? `(${evolutionLevel})` : ''} → ${evoChain} **(${poke.evolutionLevel})**`;
+
+				// If the direct pre-evolution has another pre-evolution (charizard -> charmeleon -> charmander)
+				if (poke.preevolutions[0].preevolutions?.length) {
+					evoChain = `\`${toTitleCase(poke.preevolutions[0].preevolutions[0].species)}\` → ${evoChain}`;
+				}
+			}
```
kyranet: This can be refactored, preferably with a helper function.

Essentially, you can do a single `evoChain` assignment instead of 2 (3 with the initial set).

src/commands/Pokemon/pokedex.ts
```diff
@@ -0,0 +1,176 @@
...
+				const otherFormeEvos = poke.evolutions.slice(1);
+				if (otherFormeEvos.length) {
+					evoChain = `${evoChain}, ${otherFormeEvos.map(oevo => `\`${oevo.species}\` (${oevo.evolutionLevel})`).join(', ')}`;
+				}
+
+				// If the direct evolution has another evolution (charmander -> charmeleon -> charizard)
+				if (poke.evolutions[0].evolutions?.length) {
+					evoChain = `${evoChain} → \`${toTitleCase(poke.evolutions[0].evolutions[0].species)}\` (${poke.evolutions[0].evolutions[0].evolutionLevel})`;
+				}
+			}
```
kyranet: Same thing as above.

src/commands/Pokemon/pokedex.ts
```diff
@@ -0,0 +1,176 @@
...
+import { fetch } from '../../lib/util/util';
+
+const BaseStats: Record<string, string> = {
+	hp: 'HP',
+	attack: 'ATK',
+	defense: 'DEF',
+	specialattack: 'SPA',
+	specialdefense: 'SPD',
+	speed: 'SPE'
+};
```
kyranet: Any reason why this is not an `enum`?

src/commands/Pokemon/pokedex.ts
```diff
@@ -0,0 +1,176 @@
...
+			description: language => language.tget('COMMAND_POKEDEX_DESCRIPTION'),
+			extendedHelp: language => language.tget('COMMAND_POKEDEX_EXTENDED'),
+			requiredPermissions: ['EMBED_LINKS'],
+			usage: '<pokemon:str>'
+		});
+	}
+
+	public async run(message: KlasaMessage, [pokemon]: [string]) {
+		try {
+			const embedTranslations = message.language.tget('COMMAND_POKEDEX_EMBED_DATA');
```
kyranet: This should be declared the moment before it's used, not in the very start of the function, same for all other variables/constants.

kyranet (changes requested):

src/languages/en-US.ts
```diff
@@ -1466,6 +1474,28 @@ export default class extends Language {
...
+			unknownSmogonTier: 'Unknown / Alt form',
+			height: 'Height',
+			weight: 'Width',
+			eggGroups: 'Egg group(s)',
+			otherFormes: 'Other forme(s)',
+			evolutionaryLine: 'Evolutionary line',
+			baseStats: 'Base stats',
+			flavourText: 'Pokdex entry',
+			externalResources: 'External resources',
+			none: 'None'
```
kyranet: The convention is that subkeys should be in SCREAMING_CASE as well, likewise all other keys.

kyranet (changes requested):

src/commands/Pokemon/pokedex.ts
```diff
@@ -0,0 +1,169 @@
...
+
+	public constructor(store: CommandStore, file: string[], directory: string) {
+		super(store, file, directory, {
+			aliases: ['pokemon', 'dex'],
+			cooldown: 10,
+			description: language => language.tget('COMMAND_POKEDEX_DESCRIPTION'),
+			extendedHelp: language => language.tget('COMMAND_POKEDEX_EXTENDED'),
+			requiredPermissions: ['EMBED_LINKS'],
+			usage: '<pokemon:pokemon>',
+			flagSupport: true
```
PyroTechniac, line 28: Not needed, flagSupport is enabled by default

src/languages/es-ES.ts
```diff
@@ -1444,6 +1456,28 @@ export default class extends Language {
 			examples: ['@Pete Attempted to mention everyone.']
 		}),
 
+		/**
+		 * ##################
+		 * Pokémon COMMANDS
```
PyroTechniac: Pokemon should be all caps

src/commands/Pokemon/pokedex.ts
```diff
@@ -0,0 +1,169 @@
...
+		});
+
+		this.createCustomResolver('pokemon', arg => arg.toLowerCase());
```
PyroTechniac: No need to create a custom resolver, just cast it to lower case in the run method

src/commands/Pokemon/pokedex.ts
```diff
@@ -0,0 +1,169 @@
...
+				body: JSON.stringify({
+					query: getPokemonDetailsByFuzzy(pokemon)
+				})
+			}, 'json') as Promise<GraphQLPokemonResponse<'getPokemonDetailsByFuzzy'>>;
+		} catch (err) {
+			throw message.language.tget('SYSTEM_QUERY_FAIL');
+		}
+	}
+
+	private resolveColour(col: string) {
```
PyroTechniac, line 125: This could probably use an enum in case we use it later

PyroTechniac (changes requested):

src/commands/Pokemon/pokedex.ts
kyranet: It's actually not.

src/commands/Pokemon/pokedex.ts
favna: The colours it switch-cases are the colours are officially defined by the Pokedex so I disagree.

PyroTechniac: I meant a string enum for colors

favna: Can't be done. Given this enum:
```ts
enum PokemonColours {
	'Black' = 0x323232,
	'Blue' = 0x257CFF,
	'Brown' = 0xA3501A,
	'Gray' = 0x969696,
	'Green' = 0x3EFF4E,
	'Pink' = 0xFF65A5,
	'Purple' = 0xA63DE8,
	'Red' = 0xFF3232,
	'White' = 0xE1E1E1,
	'Yellow' = 0xFFF359,
}
```

Results in this error:
image


Making it `const enum` is no good either because const enums can only be accessed through dot notation.
Casting `poke.colors` is impossible as well.

kyranet (approved):

PR #510, opened 2019-11-13
docs: update contributing guidelines

favna:
Rewrote some parts so the sentences flow a bit better and added a section about what has to be edited in the config to run Skyra locally. Or in other words, written from the perspective of someone who doesn't know the entire codebase and systems. 

p.s. The Queen's English is the only English. It's `behaviour`, not that murican lame version of `behavior` 😏


kyranet (approved):
I simply copypasted Dirigeants's, but yeah, thanks you!

PR #512, opened 2019-11-13
feat: Refactor usernames and nicknames for a better system

kyranet:
This new system also helps querying the API less!

### DIFF
--- a/src/arguments/membername.ts
+++ b/src/arguments/membername.ts
@@ -10,7 +10,7 @@ export default class extends Argument {
 		const resMember = await this.resolveMember(message, arg);
 		if (resMember) return resMember;
 
-		const result = await new FuzzySearch(message.guild!.memberUsernames, entry => entry, filter).run(message, arg, possible.min || undefined);
+		const result = await new FuzzySearch(message.guild!.nicknames.mapUsernames(), entry => entry, filter).run(message, arg, possible.min || undefined);
 		if (result) {
 			return message.guild!.members.fetch(result[0])
 				.catch(() => { throw message.language.tget('USER_NOT_EXISTENT'); });
@@ -22,7 +22,7 @@ export default class extends Argument {
 		const id = USER_REGEXP.test(query)
 			? USER_REGEXP.exec(query)![1]
 			: USER_TAG.test(query)
-				? this.client.usertags.findKey(tag => tag === query) || null
+				? this.client.userTags.getKeyFromTag(query) || null
 				: null;
 
 		if (id) {
--- a/src/commands/Tools/poll.ts
+++ b/src/commands/Tools/poll.ts
@@ -141,10 +141,10 @@ export default class extends SkyraCommand {
 	public async _resolveUsers(message: KlasaMessage, users: string[]) {
 		const output: string[] = [];
 		for (const user of users) {
-			let resolved: string | undefined;
+			let resolved: string | null | undefined;
 			if (REG_USERS.test(user)) resolved = (await message.guild!.members.fetch(REG_USERS.exec(user)![1])).id;
-			else if (REG_TAG.test(user)) resolved = message.guild!.memberTags.findKey(tag => tag === user);
-			else resolved = message.guild!.memberUsernames.findKey(tag => tag === user);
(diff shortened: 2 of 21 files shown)


cfanoulis (dismissed):
LGTM (at least the code I could understand)
minor nit I just thought about just as I left this: a helper to return the username#tag to skip the template literals?

src/arguments/membername.ts
```diff
@@ -10,7 +10,7 @@ export default class extends Argument {
 		const resMember = await this.resolveMember(message, arg);
 		if (resMember) return resMember;
 
-		const result = await new FuzzySearch(message.guild!.memberUsernames, entry => entry, filter).run(message, arg, possible.min || undefined);
+		const result = await new FuzzySearch(message.guild!.nicknames.mapUsernames(), entry => entry, filter).run(message, arg, possible.min || undefined);
```
favna, line 13: I thought DJS fixed the need for the non null assertion for message.guild? Same for username if so.

src/commands/Tools/poll.ts
```diff
@@ -141,10 +141,10 @@ export default class extends SkyraCommand {
 	public async _resolveUsers(message: KlasaMessage, users: string[]) {
 		const output: string[] = [];
 		for (const user of users) {
-			let resolved: string | undefined;
+			let resolved: string | null | undefined;
```
favna, line 144: What's the point of `null | undefined` when you're not checking specifically `=== null` or `=== undefined` afterwards?

src/lib/util/FuzzySearch.ts
```diff
@@ -1,6 +1,7 @@
-import { Collection, Message } from 'discord.js';
+import { Message } from 'discord.js';
 import { util } from 'klasa';
 import { levenshtein } from './External/levenshtein';
+import { Collection } from '@discordjs/collection';
```
favna, line 4: Why not import from discord.js?

favna (commented):
Few more remarks

src/arguments/membername.ts
kyranet: https://github.com/kyranet/Skyra/blob/4cb98c2aa110d0ef4ae2aa062920acf27705635d/src/commands/Tools/poll.ts#L147

Returns `string | null`.

src/commands/Tools/poll.ts
kyranet: https://github.com/kyranet/Skyra/blob/4cb98c2aa110d0ef4ae2aa062920acf27705635d/src/commands/Tools/poll.ts#L147

Returns `string | null`.

src/arguments/membername.ts
kyranet: Messages sent in a DM channel have `guild === null`.

src/lib/util/FuzzySearch.ts
kyranet: discord.js's Collection has `toJSON()`, which we don't need.

vladfrangu (changes requested):
A thing I noticed

src/lib/util/util.ts
```diff
@@ -294,6 +295,14 @@ export function getColor(message: Message) {
 	return message.author.settings.get(UserSettings.Color) || (message.member && message.member.displayColor) || BrandingColors.Primary;
 }
 
+const ROOT = 'https://cdn.discordapp.com';
```
vladfrangu, line 298: This is wrong. Iirc its discriminator % 5 

Good job FastHub... Discriminator line if the user has no avatar

kyranet: Fixed in c9c71c1

src/lib/util/FuzzySearch.ts
favna: Yes but for consistency in the code I would think using the discord.js one would be better. Surely having that method and not even calling it has no real overhead?

kyranet: It's unnecessary, that's why.

And as a general rule, I avoid all of discord.js's `toJSON` at all costs.

It gives all the data serialized plus bonus useless stuff.

favna (approved):

PR #519, opened 2019-11-14
Fix(Markov): Fix caching for users

### LINKED ISSUE #518: bug: Markov caches user model when using markov with a user
**Describe the bug**

Essentially, when running markov, Skyra caches the trained model for 1 minute, this is problematic when using the user filter, because the model will be for said user for that time.

**To Reproduce**
1. Write `s!markov @kyra`
2. Write `s!markov`
3. All messages from the second command will be filtered for the user

**Expected behavior**

If markov is run with filter, it should have its own cache.

**Screenshots**
None.

**Additional context**
None.

PyroTechniac:
Should fix #518

### DIFF
--- a/src/commands/Fun/markov.ts
+++ b/src/commands/Fun/markov.ts
@@ -13,6 +13,7 @@ export default class extends SkyraCommand {
 
 	private readonly kMessageHundredsLimit = 10;
 	private readonly kInternalCache = new WeakMap<TextChannel, Markov>();
+	private readonly kInternalUserCache = new Map<string, Markov>();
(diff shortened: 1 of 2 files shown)


kyranet (changes requested):
Now the bug this introduced: A user model is trained in one channel, what happens when in another channel, a user runs the markov against the same user?

PyroTechniac:
> Now the bug this introduced: A user model is trained in one channel, what happens when in another channel, a user runs the markov against the same user?

Hmm, maybe keying the user-trained markovs by `channelID.userID` would be a good solution

kyranet (approved):

PR #535, opened 2019-11-17
feat: Added restriction commands

kyranet:


src/commands/Moderation/Utilities/time.ts
```diff
@@ -96,6 +96,7 @@ export default class extends SkyraCommand {
 			case Moderation.TypeCodes.FastTemporaryVoiceMute:
 			case Moderation.TypeCodes.TemporaryVoiceMute:
 			case Moderation.TypeCodes.VoiceMute: return this.checkVMute(message, user);
+			// TODO: Add the rest of restrictions
```
favna, line 99: Is this PR still a WIP?

src/languages/es-ES.ts
```diff
@@ -1448,6 +1448,42 @@ export default class extends Language {
...
+					a temporary restriction. This action saves a member's roles temporarily and will be granted to the user after
+					the reverted restriction.
+					The reaction role is **sticky**, if the user tries to remove it by rejoining the guild, it will be added back.`,
+			examples: ['@Pete', '@Pete Spamming reactions', '@Pete 24h Posting weird reactions']
+		}),
+		COMMAND_RESTRICTVOICE_DESCRIPTION: 'Restrict a user from joining any voice channel.',
+		COMMAND_RESTRICTVOICE_EXTENDED: builder.display('restrictVoice', {
+			extendedHelp: `This command requires **${PERMS.MANAGE_ROLES}**, and only members with lower role hierarchy position
+					can be managed by me. No, the guild's owner cannot be restricted. This action can be optionally timed to create
+					a temporary restriction. This action saves a member's roles temporarily and will be granted to the user after
```
favna, line 1482: CC: https://github.com/kyranet/Skyra/issues/546

Also all other lines in es-ES

Just resolve it when seen. The reference is on the issue now.

src/commands/Moderation/Utilities/time.ts
kyranet: Technically, yes, since this is missing. But it should be simple enough for anybody else to pick it up, which is why I didn't do `// TODO(kyranet)`.

I want to work on other fixes and this is too tedious.

src/languages/es-ES.ts
kyranet: Fixed with 4694fad.

favna (approved):

PR #547, opened 2019-11-18
refactor(util.fetch): Use const enum instead of string literals

Quantumlyy:


src/lib/util/util.ts
```diff
@@ -215,42 +215,49 @@ export async function fetchAllLeaderboardEntries(client: Client, results: readon
...
+export async function fetch(url: URL | string, options: RequestInit, type: FetchResultTypes.Text): Promise<string>;
+export async function fetch(url: URL | string, type: FetchResultTypes.Result): Promise<Response>;
+export async function fetch(url: URL | string, options: RequestInit, type: FetchResultTypes.Result): Promise<Response>;
+export async function fetch(url: URL | string, options: RequestInit, type: FetchResultTypes): Promise<Response | Buffer | string | unknown>;
+export async function fetch(url: URL | string, options: RequestInit | FetchResultTypes, type?: FetchResultTypes) {
 	if (typeof options === 'undefined') {
 		options = {};
-		type = 'json';
+		type = FetchResultTypes.JSON;
 	} else if (typeof options === 'string') {
```
kyranet: By default, enum values are numbers:
```suggestion
	} else if (typeof options === 'number') {
```

kyranet (changes requested):

favna (approved):
LGTM

kyranet (approved):
Should work™️, and the code is now more readable. Thanks!

PR #553, opened 2019-11-18
Feat(Core): Twitch stream notifications

### LINKED ISSUE #318: Feat(Core): Twitch Streams.

kyranet:
Fixes #318


src/lib/util/Notifications/Twitch.ts
```diff
@@ -0,0 +1,91 @@
...
+		replaceEncode ? out = out.replace(/%20/g, '&20') : null;
+		return out;
+	}
+
+	private async _performApiGETRequest<T>(path: string, api: ApiVersion = ApiVersion.Kraken): Promise<T> {
+		const result = await fetch(`${api === ApiVersion.Kraken ? this.BASE_URL_KRAKEN : this.BASE_URL_HELIX}${path}`, this.kFetchOptions, FetchResultTypes.JSON) as unknown as T;
+		return result;
+	}
+
+	public static RATELIMIT_COOLDOWN: number = 10 * 60 * 1000;
```
PyroTechniac: This should use the Time enum found in `lib/util/constants.ts` and multiply `Time.Minute` by 10

PyroTechniac (changes requested):

src/lib/util/Notifications/Twitch.ts
kyranet: Resolved with 38cb96f.

PyroTechniac (approved):

src/commands/Twitch/twitchsubscription.ts
```diff
@@ -0,0 +1,257 @@
...
+import { BrandingColors } from '../../lib/util/constants';
+
+const enum Type {
+	Add = 'add',
+	Remove = 'remove',
+	Reset = 'reset',
+	Show = 'show'
+}
+
+type Streamer = TwitchHelixUsersSearchResult;
```
vladfrangu, line 21: Why are all of these aliases a thing?

src/lib/util/Notifications/Twitch.ts
```diff
@@ -0,0 +1,149 @@
...
+	access_token: string;
+	refresh_token: string;
+	scope: string;
+	expires_in: number;
+}
+
+export class Twitch {
+
+	public readonly ratelimitsStreams = new RateLimitManager(1, Twitch.RATELIMIT_COOLDOWN);
+	public readonly BASE_URL_HELIX = 'https://api.twitch.tv/helix/';
```
vladfrangu, line 28: Shouldn't this use constants?

vladfrangu (changes requested):

src/commands/Twitch/twitchsubscription.ts
kyranet: For readability and for shortening the lines in the command arguments, it's hard to read them when they reach +200 characters.

src/lib/util/Notifications/Twitch.ts
kyranet: While most of the code uses constants, the newest code has been migrating to moving them as readonly fields, which makes them easier to inspect when debugging or even, to modify (with eval).

This also is, because it makes more sense for them to be class fields when dealing with pure OOP, improving the readability of the code.

PR #561, opened 2019-11-19
Fix(Events): Unload once events

PyroTechniac:
Technically gives a small performance boost as well.

### DIFF
--- a/src/events/readySweeper.ts
+++ b/src/events/readySweeper.ts
@@ -6,7 +6,7 @@ const HEADER = `\u001B[39m\u001B[94m[MEMORY CLEANUP]\u001B[39m\u001B[90m`;
 export default class extends Event {
 
 	public constructor(store: EventStore, file: string[], directory: string) {
-		super(store, file, directory, { event: 'ready' });
+		super(store, file, directory, { event: 'ready', once: true });
 	}
 
 	public run() {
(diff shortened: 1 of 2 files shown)


favna (approved):
Every little bit matters 👌

src/events/readySweeper.ts
```diff
@@ -6,7 +6,7 @@ const HEADER = `\u001B[39m\u001B[94m[MEMORY CLEANUP]\u001B[39m\u001B[90m`;
 export default class extends Event {
 
 	public constructor(store: EventStore, file: string[], directory: string) {
-		super(store, file, directory, { event: 'ready' });
```
cfanoulis, line 9: I... don't think that runs only once

cfanoulis (changes requested):
Hmmm... something's not quite right

src/events/readySweeper.ts
PyroTechniac: It's for the `ready` event, which is only fired once

favna: Ready is only triggered once, specifically, when the bot is ready. It is triggered immediately before KlasaReady.

cfanoulis: whoops, didn't see that

/resolved

cfanoulis (approved):
surprised this isn't a thing already :omegaLOL:

kyranet (commented):
Nice catch, thanks you!

kyranet (approved):

PR #568, opened 2019-11-23
feat: add pokemon items command

favna:
This PR will be adding various new Pokemon commands. Writing them up ahead of time in development until the gen8 data is available because I'm also adding necessary fields

Don't look at the `as any` parts in items for example - that'll be fixed with the gen8 data release of the API.


favna:
superseded by #595

PR #579, opened 2019-11-25
add post route for syncing user

gc:
Adds a route for users to update their data

Ratelimit is 2 per 5 mins

### DIFF
--- a/src/routes/oauth/oauthUser.ts
+++ b/src/routes/oauth/oauthUser.ts
@@ -4,6 +4,7 @@ import { ratelimit, authenticated } from '../../lib/util/util';
 import ApiRequest from '../../lib/structures/api/ApiRequest';
 import ApiResponse from '../../lib/structures/api/ApiResponse';
 import { Events } from '../../lib/types/Enums';
+import { Time } from '../../lib/util/constants';
 
 export default class extends Route {
 
@@ -22,7 +23,7 @@ export default class extends Route {
 	}
 
 	@authenticated
-	@ratelimit(2, 1000 * 60 * 5, true)
+	@ratelimit(2, Time.Minute * 5, true)
 	public async post(request: ApiRequest, response: ApiResponse) {
 		const requestBody = request.body as Record<string, string>;
 		if (!requestBody.action) {
--- a/src/routes/oauth/oauthUser.ts
+++ b/src/routes/oauth/oauthUser.ts
@@ -35,11 +35,7 @@ export default class extends Route {
 					this.client.emit(Events.Wtf, err);
 				});
 
-			if (user) {
-				 return response.json({ user });
-			}
-
-			return response.error(500);
+			return user ? response.json({ user }) : response.error(500);
 		}
 
 		return response.error(400);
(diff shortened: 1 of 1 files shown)


src/routes/oauth/oauthUser.ts
```diff
@@ -17,4 +21,24 @@ export default class extends Route {
 		return this.client.dashboardUsers.add(user);
 	}
 
+	@authenticated
+	@ratelimit(2, 1000 * 60 * 5, true)
```
PyroTechniac:
```suggestion
	@ratelimit(2, 300000, true)
```
Should use constant number for consistency

src/routes/oauth/oauthUser.ts
```diff
@@ -17,4 +21,24 @@ export default class extends Route {
...
+		if (requestBody.action === 'SYNC_USER') {
+			const user = await this.api(request.auth!.token)
+				.catch(err => {
+					this.client.emit(Events.Wtf, err);
+				});
+
+			return user ? response.json({ user }) : response.error(500);
+		}
+
+		return response.error(400);
```
PyroTechniac: This could all be changed to check if the action is not `SYNC_USER`, and error accordingly

PyroTechniac (changes requested):

src/routes/oauth/oauthUser.ts
gc: It's coded to be easy to add more actions

kyranet: This can also error because the token expires in a week, we need to save the refresh and the expire at values somewhere.

My idea was making a table for that, it's not suitable in user settings due to it making the db and RAM usage much bigger.

src/routes/oauth/oauthUser.ts
```diff
@@ -17,4 +22,24 @@ export default class extends Route {
 		return this.client.dashboardUsers.add(user);
 	}
 
+	@authenticated
+	@ratelimit(2, Time.Minute * 5, true)
+	public async post(request: ApiRequest, response: ApiResponse) {
+		const requestBody = request.body as Record<string, string>;
+		if (!requestBody.action) {
```
kyranet:
```suggestion
		if (typeof requestBody.action !== 'string') {
```

src/routes/oauth/oauthUser.ts
```diff
@@ -17,4 +22,24 @@ export default class extends Route {
...
+			return response.error(400);
+		}
+
+		if (requestBody.action === 'SYNC_USER') {
+			const user = await this.api(request.auth!.token)
+				.catch(err => {
+					this.client.emit(Events.Wtf, err);
+				});
+
+			return user ? response.json({ user }) : response.error(500);
```
kyranet:
```typescript
try {
	const user = await this.api(request.auth!.token);
	return response.json({ user });
} catch (error) {
	this.client.emit(Events.Wtf, error);
	return response.error(500);
}
```

kyranet (changes requested):

kyranet:
Superseded by #599

PR #588, opened 2019-11-26
feat: Added filters and position

### LINKED ISSUE #585: request: Prune after
`.prune after <message_id>`
deletes all messages *after* that message, very useful in a lot of situations

Not sure how it works if the message is very old with thousands of messages after it, maybe it limits itself to X bulk deletes before it gives up.

kyranet:
Fixes #585


src/commands/Moderation/prune.ts
```diff
@@ -20,59 +51,102 @@ export default class extends SkyraCommand {
...
 			requiredPermissions: ['MANAGE_MESSAGES', 'READ_MESSAGE_HISTORY'],
 			runIn: ['text'],
-			usage: '[limit:integer{1,100}] [link|invite|bots|you|me|upload|user:user]',
+			usage: '[limit:integer{1,100}] (position:position) (message:message) [filter:filter|user:user]',
 			usageDelim: ' '
 		});
+
+		this.createCustomResolver('position', (argument, _possible, message) => {
+			if (!argument) return null;
+			const position = this.kPositions.get(argument.toLowerCase());
```
vladfrangu: Can't we technically i18n this?

src/commands/Moderation/prune.ts
```diff
@@ -20,59 +51,102 @@ export default class extends SkyraCommand {
...
+			if (typeof position === 'undefined') throw message.language.tget('COMMAND_PRUNE_INVALID_POSITION');
+			return position;
+		}).createCustomResolver('message', async (argument, possible, message, [, position]) => {
+			if (position === null) return message;
+
+			const fetched = this.kMessageRegExp.test(argument) ? await message.channel.messages.fetch(argument).catch(() => null) : null;
```
vladfrangu, line 37: You could also add support for message links, but that might require a regex change in klasa core

src/commands/Moderation/prune.ts
```diff
@@ -20,59 +51,102 @@ export default class extends SkyraCommand {
...
+			if (fetched === null) throw message.language.tget('RESOLVER_INVALID_MESSAGE', possible.name);
+			return fetched;
+		}).createCustomResolver('filter', (argument, _possible, message) => {
+			if (!argument) return undefined;
+			const filter = this.kFilters.get(argument.toLowerCase());
```
vladfrangu: Ditto as position

src/commands/Moderation/prune.ts
```diff
@@ -20,59 +51,102 @@ export default class extends SkyraCommand {
...
 
-		let messages = await message.channel.messages.fetch({ limit: 100, before: message.id });
-		if (typeof filter !== 'undefined') {
-			const user = typeof filter === 'string' ? null : filter;
-			const type = typeof filter === 'string' ? filter : 'user';
-			messages = messages.filter(this.getFilter(message, type, user));
+		const position = this.resolvePosition(rawPosition);
+		const filter = this.resolveFilter(rawFilter);
+
+		this.client.console.debug(limit, position, targetMessage.id, filter);
```
vladfrangu: You can probably remove this now

vladfrangu (changes requested):

src/commands/Moderation/prune.ts
kyranet: No, it leaves margin for many errors, specially when it points to another channel.

vladfrangu: Aaand you can then check if the channel ID matches the current channel? 
```ts
const match = argument.match(/https?:\/\/(?:(ptb|canary)\.)?discordapp\.com\/channels\/\d+\/(\d+)\/(\d+)/);
if (match) {
	const [channelID, messageID] = match;
	if (channelID !== message.channel.id) throw 'bad';
}
```
Doesn't seem TOO trivial. This is, again, just a suggestion, not a must do 😄

kyranet: You can make a PR later adding this, but for now, I want to get this PR finished and landed.

vladfrangu (approved):

PR #592, opened 2019-11-26
fix: dev/prod environment setup

favna:
Set the URL of the API to the actual URL. At least best solution until config is rewritten as Kyra let known would happen.


kyranet:
Closed in favor of #593

PR #602, opened 2019-11-28
feat(Influx): Add influx base

Quantumlyy:
A base PR for Influx such as making the first class that interacts with it as an example.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -24,10 +24,9 @@
 /src/lib/util/Notifications/ @quantumlytangled
 /src/lib/types/settings/raw/RawTwitchStreamSubscriptionSettings.ts @quantumlytangled
 /src/lib/types/definitions/Twitch.d.ts @quantumlytangled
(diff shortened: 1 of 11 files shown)


.github/CODEOWNERS
```diff
@@ -24,6 +24,7 @@
 /src/lib/util/Notifications/ @quantumlytangled
 /src/lib/types/settings/raw/RawTwitchStreamSubscriptionSettings.ts @quantumlytangled
 /src/lib/types/definitions/Twitch.d.ts @quantumlytangled
+/docker/ @quantumlytangled
```
PyroTechniac:
```suggestion
/docker/influxdb/ @quantumlytangled
```

PyroTechniac (changes requested):

.github/CODEOWNERS
PyroTechniac: Alternatively you could leave a comment as to why you marked yourself as a codeowner of the whole `docker` folder

PR #607, opened 2019-12-01
base(Analytics): Influx Analytics Base.

### LINKED ISSUE #606: request: audit logs
Log changes to skyras settings (anything that people might have a need to see, like discord audit logs). Audit logs would be undeletable, and expire after a while (although it would be cool if it was permanent)

 My initial idea of this is something like:

```js
{
   date: new Date().getTime(),
   user_id: 455664456633,
   action: "CHANGE_SETTING",
   data: {
      setting: "channels.imagelogs",
      newValue: 468446446433
   }
}
```
We could then have something like a human readable name map to display it, e.g. that setting string maps to "Image Logs Channel"

Note that this also would need to support the dashboard, and how that updates settings. It would then ofcourse be viewable on the dashboard and perhaps a command to view the audit logs.

Would also log all punishment stuff, as mods can delete those staff logs messages.

Quantumlyy:


src/commands/Announcement/announcement.ts
```diff
@@ -68,16 +70,41 @@ export default class extends SkyraCommand {
...
+				},
+				tags: {
+					shard: (this.client.options.shards as number[])[0].toString(),
+					user_id: message.author.id,
+					guild_id: message.guild?.id!,
+					channel_id: channel.id
+				}
+			}
+		]);
+	}
```
kyranet: To avoid injecting InfluxDB code within pieces (which we don't even do with other stuff), you should make an event called `guildAnnouncementSuccess` and `guildAnnouncementError`, then move the InfluxDB logic there.

kyranet (changes requested):

gc:
Couple things: settingsUpdate logs need to include the user_id, and they shouldn't store the old_value, to get the old value you would Just find the previous entry for that key

QuantumlyTangled:
V

kyranet (approved):

PR #610, opened 2019-12-02
feat: topinvites command

### LINKED ISSUE #534: Top Invites command
ordos implementation:
```js
const invites = await msg.guild.fetchInvites();
const topTen = invites.filter(inv => inv.uses > 0).sort((a, b) => b.uses - a.uses).first(10);
if (topTen.length === 0) throw 'There are no invites, or none of them have been used!';
return msg.sendMessage(
    topTen.map(inv => `**${inv.inviter.username}**'s invite **${inv.code}** has **${inv.uses.toLocaleString()}** uses.`)
);
```
image

favna:
Fixes #534 

Preview: https://streamable.com/pwijx


src/commands/Tools/topinvites.ts
```diff
@@ -0,0 +1,34 @@
...
+			description: language => language.tget('COMMAND_TOPINVITES_DESCRIPTION'),
+			extendedHelp: language => language.tget('COMMAND_TOPINVITES_EXTENDED'),
+			requiredGuildPermissions: ['MANAGE_GUILD']
+		});
+	}
+
+	public async run(message: KlasaMessage) {
+		const invites = await message.guild!.fetchInvites();
+		const topTen = invites
+			.filter(inv => this.filterInvites(inv))
```
gc: `.filter(this.filterInvites)`

kyranet (changes requested):
1. URL can be omitted and displayed as a masked url, e.g. **[`DAeQgE`](https://discord.gg/DAeQgE)**.
1. Date created and Expiry date could use a more compact format, e.g. **2019/12/3 00:04**.
1. Expiry date and maximum age can be shortened in one, e.g. **2019/12/3 01:04 (1h)**.
1. Maximum uses and amount of uses can be turned into one, e.g. **5**, or **5/10** (when limit is 10).

This would reduce the amount of fields from 9 to 6, removing one entire row while reducing verbosity.

Furthermore you can also use the description to format the uses (with limit if defined), channel, code (with masked url), and whether it's temporary or not. And then use two fields to format the date, e.g:

```markdown
**Uses**: 5/10
**Link**: [`DAeQgE`](https://discord.gg/DAeQgE)
**Channel**: #bot-spam
**Temporary**: ✔️ 

Date Created      Expiry Date
2019/12/3 00:04   2019/12/3 01:04 (1h)
```

src/commands/Tools/topinvites.ts
PyroTechniac: Actually, this would require binding of the method due to `this` becoming undefined with an unbound method

src/commands/Tools/topinvites.ts
```diff
@@ -0,0 +1,72 @@
...
+
+		if (topTen.length === 0) throw message.language.tget('COMMAND_TOPINVITES_NO_INVITES');
+
+		const display = this.buildDisplay(message, topTen);
+
+		await display.start(response, message.author.id);
+		return response;
+	}
+
+	private filterInvites(invite: Invite) {
```
PyroTechniac: This method could be made into simply
```ts
return Boolean(invite.uses && invite.inviter?.username)
```

src/commands/Tools/topinvites.ts
```diff
@@ -0,0 +1,72 @@
...
+		if (Boolean(invite.uses) && invite.inviter && invite.inviter.username) return true;
+		return false;
+	}
+
+	private buildDisplay(message: KlasaMessage, invites: DeepRequired<Invite[]>) {
+		const display = new UserRichDisplay(new MessageEmbed()
+			.setTitle(`Top 10 Invites for ${message.guild!.name}`)
```
PyroTechniac: This should use a language key

PyroTechniac (changes requested):

src/commands/Tools/topinvites.ts
PyroTechniac: Or, better yet
```ts
return (invite.uses && invite.inviter?.username)
```

kyranet: I dislike the latter, plus filter expects a boolean, so it'll be better performance-wise to always give it a boolean.

Also, the username check is unnecessary, user objects are never partial, this should be:

```typescript
return invite.uses > 0 && invite.inviter !== null;
```

PyroTechniac: Couldn't that be simplified to 
```typescript
Boolean(invite.uses && invite.inviter)
```
as both `0` and `null` are falsy?

kyranet: It's unnecessary overhead and a failure here means a type mismatch, which will (and should) error afterwards rather than failing silently.

This check gives a guarantee of pure type safety.

src/commands/Tools/topinvites.ts
gc: so bind it? 🤣

favna: Exactly what Pyro said, that's what I wrote at first. @kyranet's call on what he prefers but considering he upvoted Pyro's comment I assume he prefers the current way.

src/commands/Tools/topinvites.ts
favna: Will do

src/commands/Tools/topinvites.ts
favna: So... what I wrote is correct then?

kyranet: No, you wrote this:
https://github.com/kyranet/Skyra/blob/436526640f0a25cf41d7e5ed5bf0ea528b356202/src/commands/Tools/topinvites.ts#L43-L46

src/commands/Tools/topinvites.ts
```diff
@@ -0,0 +1,72 @@
...
+		if (topTen.length === 0) throw message.language.tget('COMMAND_TOPINVITES_NO_INVITES');
+
+		const display = this.buildDisplay(message, topTen);
+
+		await display.start(response, message.author.id);
+		return response;
+	}
+
+	private filterInvites(invite: Invite) {
+		if (Boolean(invite.uses) && invite.inviter && invite.inviter.username) return true;
```
favna:
```suggestion
```

src/commands/Tools/topinvites.ts
```diff
@@ -0,0 +1,72 @@
...
+		return false;
```
favna:
```suggestion
		return invite.uses > 0 && invite.inviter !== null;
```

favna (commented):
lazy mode active

src/commands/Tools/topinvites.ts
kyranet: I prefer bind, but I'm fine with this one as well.

src/commands/Tools/topinvites.ts
```diff
@@ -0,0 +1,71 @@
...
+		const display = new UserRichDisplay(new MessageEmbed()
+			.setTitle(`Top 10 Invites for ${message.guild!.name}`)
+			.setColor(getColor(message)));
+		const embedData = message.language.tget('COMMAND_TOPINVITES_EMBED_DATA');
+
+		for (const invite of invites) {
+			display.addPage((embed: MessageEmbed) => embed
+				.setAuthor(invite.inviter.tag, invite.inviter.displayAvatarURL())
+				.setThumbnail(invite.inviter.displayAvatarURL())
+				.addField(embedData.CHANNEL, `<#${invite.channel.id}>`, true)
```
PyroTechniac:
```suggestion
				.addField(embedData.CHANNEL, `${invite.channel}`, true)
```
You don't have to manually format the channel, you can just let JavaScript do it

PyroTechniac (dismissed):

src/commands/Tools/topinvites.ts
favna: nvm

PyroTechniac: Yes, and channels have a toString method that does the same thing you're doing, so why not let JavaScript do it instead of manually formatting it?

kyranet (dismissed):

src/commands/Tools/topinvites.ts
```diff
@@ -0,0 +1,92 @@
...
+import { BrandingColors, Emojis } from '../../lib/util/constants';
+import { getColor } from '../../lib/util/util';
+
+export default class extends SkyraCommand {
+
+	private inviteTimestamp = new Timestamp('YYYY/MM/DD HH:mm');
+	private filter: (invite: Invite) => boolean;
+
+	public constructor(store: CommandStore, file: string[], directory: string) {
+
```
kyranet:
```suggestion
```

kyranet (changes requested):

kyranet (approved):

PR #618, opened 2019-12-04
security: ensure inaccessible channels cannot be accessed by commands

favna:
Basically implemented the same way as markov.

This is MVP (Minimum Viable Product) but maybe we should refactor this to be in the channelname argument for in the future? That would however mean making a `userchannel` and `adminchannel` argument as moderation commands should not have channel restrictions

### DIFF
--- a/src/commands/System/echo.ts
+++ b/src/commands/System/echo.ts
@@ -11,7 +11,7 @@ export default class extends SkyraCommand {
 			extendedHelp: language => language.tget('COMMAND_ECHO_EXTENDED'),
 			guarded: true,
 			permissionLevel: 10,
-			usage: '[channel:channelname] [message:string] [...]',
+			usage: '[channel:channelname{2}] [message:string] [...]',
 			usageDelim: ' '
 		});
 	}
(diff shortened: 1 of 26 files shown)


kyranet:
Not even admins should be able to give channels Skyra or the author cannot read, so indeed, the argument itself should take care of that, as well as ensuring the channel's guild is the message's guild.

src/commands/System/echo.ts
```diff
@@ -11,7 +11,7 @@ export default class extends SkyraCommand {
 			extendedHelp: language => language.tget('COMMAND_ECHO_EXTENDED'),
 			guarded: true,
 			permissionLevel: 10,
-			usage: '[channel:channel] [message:string] [...]',
+			usage: '[channel:channelname] [message:string] [...]',
```
kyranet: Nit: `channelname` can conflict with `string` here. Replace it with any of the following:

```suggestion
			usage: '[channel:channelname{2}] [message:string] [...]',
```

```suggestion
			usage: '[channel:channel] [message:string] [...]',
```

kyranet (changes requested):

favna:
Merged one of em but I recommend you test it with SkyraDev before pushing to production because I'm not sure if my own tests were proper.

kyranet (approved):

PR #625, opened 2019-12-05
util(Decorator): Add class method ratelimiting

Quantumlyy:


src/lib/util/Limiter.ts
```diff
@@ -0,0 +1,61 @@
...
+	limiter: boolean;
+}
+
+export interface MethodRatelimitedError {
+	error: LimitErrors.Ratelimited;
+	remainingTime: number;
+}
+
+export function classLimitInitialization() {
+	return createClassDecorator((target: Constructor<LimitedClass>) => class extends target {
```
kyranet:
```suggestion
	return createClassDecorator((Target: Constructor<unknown>) => class extends Target implements LimitedClass {
```

src/lib/util/Limiter.ts
```diff
@@ -0,0 +1,61 @@
...
+
+		public ratelimits: Collection<string, RateLimitManager> = new Collection();
```
kyranet: If the `implements` helps TypeScript inferring the type of the arguments, remove their type casts:
```suggestion
		public ratelimits = new Collection();
```

But if it doesn't:
```suggestion
		public ratelimits = new Collection<string, RateLimitManager>();
```

src/lib/util/Limiter.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export function limitMethod(group: string, bucket: number, cooldown: number, safety = true) {
+	return createMethodDecorator((_target, _propertyKey, descriptor) => {
+		const method = descriptor.value;
+
+		if (!method) throw new Error('Function limiter actions require a [[value]].');
+		if (typeof method !== 'function') throw new Error('Function limiter actions can only be applied to functions.');
+
+		descriptor.value = (function descriptorValue(this: LimitedClass, ...args: any[]) {
```
kyranet:
```suggestion
		descriptor.value = (function descriptorValue(this: LimitedClass, ...args: readonly unknown[]) {
```

src/lib/util/Limiter.ts
```diff
@@ -0,0 +1,61 @@
...
+			if (!this.limiter as boolean) throw new Error('Class does not posses a limiter');
```
kyranet: This should never happen as `this` is a `LimitedClass` instance.

src/lib/util/Limiter.ts
```diff
@@ -0,0 +1,61 @@
...
+			bucket = safety ? (bucket > 2 ? bucket - 1 : bucket) : bucket;
+			cooldown = safety ? (cooldown + (Time.Second as number)) : cooldown;
```
kyranet: I don't understand this part, and why do they have to be in the method.

src/lib/util/Limiter.ts
```diff
@@ -0,0 +1,61 @@
...
+			const manager = this.ratelimits.get(group);
+			const id = Date.now().toString(8);
+			const bucketInstance = manager!.acquire(id);
+
+			if (bucketInstance.limited) {
+				return {
+					error: LimitErrors.Ratelimited,
+					remainingTime: bucketInstance.remainingTime.toString()
+				};
+			}
```
kyranet: This smells like at least two bugs.

kyranet (changes requested):

kyranet (approved):

PR #626, opened 2019-12-06
chore: config.example.ts doesn't need to be built

cfanoulis:
`tsconfig.json` included an extra file that needn't be built. This PR removes it

~~wrong branch, I know, sosumi~~

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -15,7 +15,6 @@
 		"./src/**/*",
 		"./database/**/*",
 		"./config.ts",
-		"./config.example.ts"
 	],
 	"exclude": [
 		"**/node_modules",


kyranet (commented):
This is wrong, it's only there so ESLint runs on it as well:

image

favna:
That's a falacy if you ask me. Nothing prevents us from creating a "tsconfig.eslint.json" which extends base and is targeted in the parser options.project. It's even what the readme recommends and as a matter of fact I told Vlad and you dozens of time to use this approach. 

From the @typescript-eslint/parser readme:

> Note that if this setting is specified and createDefaultProgram is not, you must only lint files that are included in the projects as defined by the provided tsconfig.json files. If your existing configuration does not include all of the files you would like to lint, you can create a separate tsconfig.eslint.json as follows:

```json5
{
  // extend your base config so you don't have to redefine your compilerOptions
  "extends": "./tsconfig.json",
  "include": [
    "src/**/*.ts",
    "test/**/*.ts",
    "typings/**/*.ts",
    // etc
 
    // if you have a mixed JS/TS codebase, don't forget to include your JS files
    "src/**/*.js"
  ]
}
```

Adjusting the project scope (by adjusting the primary tsconfig) just to lint files is NOT the recommended method.

PR #637, opened 2019-12-07
fix: Resolved inconsistencies between routes and oauth2

### LINKED ISSUE #614: request: allow members with admin role to access dashboard
Tldr: if member has admin setting role, let them log into dashboard without needing manage server perm.

In skyra,
Mod role: allows you to use mod commands like ban and prune
Admin role: also allows you to configure settings

In other words, the admin role is the settings manager role. However you could be a skyra role admin but not have manage server perms, which means you can't manage the server on the dashboard. This is actually the case for a couple servers I'm in.

Requires:
1. Change KDH fork so DashboardGuild.userCanManage checks for the admin setting role OR the manage server perm.
2. Change whichever KDH routes need to be
3. Frontend shouldn't need any changes I believe

### LINKED ISSUE #617: bug: Inconsistency in KDH/DashboardGuild vs Guild#toJSON/Guild Route
KDH.DashboardGuild is vastly different from what we get in the Guild route, which requires ugly ugly code to fix in the frontend. If possible, it should be the exact same thing and come from the same place, so we dont have to update both if we want to add/remove something. Also means that the Guild route needs the extra props in DashboardGuild like `userCanManage`, however  `guild.iconURL` should be just `guild: guild.iconURL()`, currently it returns always a `.png.`.

Related to #614, where that issue requires we have `userIsAdmin` or something like that in the guild.

kyranet:
- [x] Fixes #617
- [x] Fixes #614
- [x] Updated skyra.pw for the new format (ref: https://github.com/kyranet/skyra.pw/pull/47)

### DIFF
--- a/src/lib/util/Models/ApiTransform.ts
+++ b/src/lib/util/Models/ApiTransform.ts
@@ -42,9 +42,9 @@ export interface FlattenedGuild {
 	splash: string | null;
 	region: string;
 	features: GuildFeatures[];
-	applicationID: string;
+	applicationID: string | null;
 	afkTimeout: number;
-	afkChannelID: string;
+	afkChannelID: string | null;
 	systemChannelID: string | null;
 	embedEnabled: boolean;
 	premiumTier: number;
--- a/src/routes/oauth/oauthCallback.ts
+++ b/src/routes/oauth/oauthCallback.ts
@@ -7,6 +7,7 @@ import OauthUser from './oauthUser';
 import { ratelimit } from '../../lib/util/util';
 import { OauthData } from '../../lib/types/DiscordAPI';
 import { Time } from '../../lib/util/constants';
+import { Events } from '../../lib/types/Enums';
 
 export default class extends Route {
 
@@ -33,7 +34,7 @@ export default class extends Route {
 		});
 
 		if (!res.ok) {
-			console.log(await res.text());
+			this.client.emit(Events.Error, `[KDH] Failed to fetch token: ${await res.text()}`);
 			response.error('There was an error fetching the token.');
 			return;
 		}
@@ -47,6 +48,7 @@ export default class extends Route {
 
 		const body = await res.json() as OauthData;
 		const user = await oauthUser.api(body.access_token);
+		if (user === null) return response.error(500);
 
 		await this.client.queries.insertDashboardUser({
 			id: user.id,
--- a/src/arguments/membername.ts
+++ b/src/arguments/membername.ts
@@ -11,7 +11,7 @@ export default class extends Argument {
 		const resMember = await this.resolveMember(message, arg);
 		if (resMember) return resMember;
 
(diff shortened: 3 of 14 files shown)


src/routes/oauth/oauthCallback.ts
```diff
@@ -47,6 +48,7 @@ export default class extends Route {
 
 		const body = await res.json() as OauthData;
 		const user = await oauthUser.api(body.access_token);
+		if (user === null) return response.error(500);
```
favna, line 51:
```suggestion
		if (user === null) return response.error(401);
```

If there is no access token in the body then the user would likely be Unauthorized, therefore a 401 should be returned.

```
    401 Unauthorized
The request requires user authentication. The response MUST include a WWW-Authenticate header field (section 14.47) containing a challenge applicable to the requested resource. The client MAY repeat the request with a suitable Authorization header field (section 14.8). If the request already included Authorization credentials, then the 401 response indicates that authorization has been refused for those credentials. If the 401 response contains the same challenge as the prior response, and the user agent has already attempted authentication at least once, then the user SHOULD be presented the entity that was given in the response, since that entity might include relevant diagnostic information. HTTP access authentication is explained in "HTTP Authentication: Basic and Digest Access Authentication".

Wikipedia
Similar to 403 Forbidden, but specifically for use when authentication is possible but has failed or not yet been provided. The response must include a WWW-Authenticate header field containing a challenge applicable to the requested resource. See Basic access authentication and Digest access authentication.

 Error code response for missing or invalid authentication token.
```

src/lib/util/Models/ApiTransform.ts
```diff
@@ -42,9 +42,9 @@ export interface FlattenedGuild {
 	splash: string | null;
 	region: string;
 	features: GuildFeatures[];
-	applicationID: string;
+	applicationID: string | null;
 	afkTimeout: number;
-	afkChannelID: string;
+	afkChannelID: string | null;
 	systemChannelID: string | null;
 	embedEnabled: boolean;
```
favna, line 49: Wouldn't it be better for the FlattenedGuild interface to extend DJS' Guild class, or if necessary, use Pick or Omit to grab those fields it is currently duplicating?

favna (commented):
Looks good so far, 2 remarks here.

src/lib/util/Models/ApiTransform.ts
kyranet: No, `FlattenGuild`'s output is strictly fixed, it doesn't care about the properties discord.js's `Guild` has.

src/routes/oauth/oauthCallback.ts
kyranet: No, because user is only `null` if it failed to fetch the user, e.g. Discord being Discord.
What is true is that I should wrap those calls within a try/catch.

gc (approved):

PR #646, opened 2019-12-10
chore: Update to SG Next

kyranet:
For some weird reason, I can't get TypeScript to recognize my augments to `@klasa/settings-gateway`, so this doesn't build.


tsconfig.json
```diff
@@ -8,7 +8,8 @@
 			"node-fetch",
 			"discord.js",
 			"klasa",
-			"klasa-dashboard-hooks"
+			"klasa-dashboard-hooks",
+			"@klasa/settings-gateway"
```
favna: Please don't. TypesSript resolves all typings in `node_modules/@types/**` and `node_modules/**/package.json -> types or typings fields`. Overwriting the automatically picking up of `@types` packages behaviour with `types` field is doing more harm than good.

https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#types-typeroots-and-types

favna (changes requested):

tsconfig.json
kyranet: What it says it does and what it actually does, is bullshit.

favna: Not a single time in the past year of writing TypeScript have I ever needed the types field. The only projects that I've seen use it are Skyra, fs-nextra and other dirigeants projects. Have you tried removing it and seeing if everything still compiles fine?

And if your claim really is true you're perfectly free to re-write it: https://github.com/microsoft/TypeScript-Website

kyranet: Of course I have, otherwise I wouldn't be able to make such assertion.
Either way this case wasn't needed, so I removed it in the following commit.

favna:
`SEM: Major`? This is going to bump Skyra to v6? Just surprised is all, since 5.3.0 is far from done.

kyranet:
No, this means that beware everyone, this can cause errors where it didn't.

Quantumlyy (approved):

PR #651, opened 2019-12-11
feat: currenttime command

favna:
This gets the current time in any place using google maps geocoding. Also standardized the same API call so there is no duplicate code between weather and currenttime

Before going into production it requires creating an account and getting a free API key from https://timezonedb.com/! Once you got your account go to the account overview to find your key.

<details>
    <summary>Screenshots (click to expand)</summary>
    <p>
        <span>
            
            
            
        </span>
    </p>
</details>


src/lib/util/Google.ts
```diff
@@ -0,0 +1,75 @@
...
+
+	if (status !== 'OK') throw message.language.tget(handleNotOK(status, client));
+	if (results.length === 0) throw message.language.tget('GOOGLE_ERROR_ZERO_RESULTS');
+
+	return [
+		results[0].formatted_address,
+		results[0].geometry.location.lat,
+		results[0].geometry.location.lng,
+		results[0].address_components
+	] as [string, number, number, GoogleMapsOkAddressComponent[]];
```
kyranet: This should return an object for readability.

kyranet (changes requested):

src/lib/util/Google.ts
favna: fixed in 42548da

kyranet (approved):

PR #652, opened 2019-12-11
feat: add emotes command

favna:
This command lists the emotes in a server, chunked by every 50 to ensure the maximum size of an embed is not exceeded (shocker but yes I had that happen with Ribbon).

Open question: 2 ways to ensure this works
1. Remove sweeping emojis from Skyra entirely
2. Keep the current implementation inside the command (lines 55-59). Advantage of this is that the cache of emojis will only hold the emojis of the servers that ran this command. 
Current implementation:
https://github.com/kyranet/Skyra/blob/4e39caeb4b11ec2e33a5a8d28ddb235349496813/src/commands/Tools/emotes.ts#L55-L59

### DIFF
--- a/src/commands/Tools/emotes.ts
+++ b/src/commands/Tools/emotes.ts
@@ -7,8 +7,6 @@ import { getColor } from '../../lib/util/util';
 
 export default class extends SkyraCommand {
 
-	private kFilter = this.removeNullAndUndefined.bind(this);
-
 	public constructor(store: CommandStore, file: string[], directory: string) {
 		super(store, file, directory, {
 			cooldown: 10,
@@ -26,8 +24,14 @@ export default class extends SkyraCommand {
 
 		await this.populateCache(message);
 
-		const animEmotes = message.guild!.emojis.map(emote => emote.animated ? `<a:${emote.name}:${emote.id}>` : undefined).filter(this.kFilter);
-		const staticEmotes = message.guild!.emojis.map(emote => emote.animated ? undefined : `<:${emote.name}:${emote.id}>`).filter(this.kFilter);
+		const animEmotes: string[] = [];
+		const staticEmotes: string[] = [];
+
+		for (const [id, emote] of [...message.guild!.emojis.entries()]) {
+			if (emote.animated) animEmotes.push(`<a:${emote.name}:${id}>`);
+			else staticEmotes.push(`<:${emote.name}:${id}>`);
+		}
+
 		const chunkedAnimatedEmojis = this.emotesChunker(animEmotes, 50);
 		const chunkedStaticEmojis = this.emotesChunker(staticEmotes, 50);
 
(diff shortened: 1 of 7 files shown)


src/commands/Tools/emotes.ts
```diff
@@ -0,0 +1,83 @@
...
+			extendedHelp: language => language.tget('COMMAND_EMOTES_EXTENDED'),
+			requiredPermissions: ['EMBED_LINKS'],
+			runIn: ['text']
+		});
+	}
+
+	public async run(message: KlasaMessage) {
+		const response = await message.sendEmbed(new MessageEmbed()
+			.setDescription(message.language.tget('SYSTEM_LOADING'))
+			.setColor(BrandingColors.Secondary));
```
cfanoulis, line 24: afaik you're supposed to use `getColor(msg)`

cfanoulis (commented):
minor nit

src/commands/Tools/emotes.ts
```diff
@@ -0,0 +1,83 @@
...
+
+		await this.populateCache(message);
+
+		const animEmotes = message.guild!.emojis.map(emote => emote.animated ? `<a:${emote.name}:${emote.id}>` : undefined).filter(this.kFilter);
+		const staticEmotes = message.guild!.emojis.map(emote => emote.animated ? undefined : `<:${emote.name}:${emote.id}>`).filter(this.kFilter);
```
gc: This does 4 loops/iterations (2 .maps and and 2 .filters), it could be done in a single `for of` loop (also agreed to by kyraa)

gc (changes requested):

src/commands/Tools/emotes.ts
favna: For the loading system message it should be brandingcolors. At least that's based on what Kyra made it like way back when I did my first work with urbandictionary soooo.....

src/commands/Tools/emotes.ts
favna: @gc updated

src/commands/Tools/emotes.ts
```diff
@@ -0,0 +1,83 @@
...
+		const chunkedAnimatedEmojis = this.emotesChunker(animEmotes, 50);
+		const chunkedStaticEmojis = this.emotesChunker(staticEmotes, 50);
+
+		const display = this.buildDisplay(message, chunkedAnimatedEmojis, chunkedStaticEmojis);
+
+		await display.start(response, message.author.id);
+		return response;
+	}
+
+	private emotesChunker(emotes: string[], size: number): string[][] {
```
gc: This should instead use the already existing `chunk` util

gc (changes requested):

kyranet (changes requested):
At this point, just remove emoji sweeping, so you don't have to fetch. The dashboard also fetches them, and once every time the guild is loaded, so soon or later I'll be ratelimited.

favna:
sweeping removed in fb8d498

kyranet (approved):

PR #663, opened 2019-12-14
feat: Websocket Infrastructure and Music dashboard

favna:


gc:
To set it up locally, run this branch (with lavalink running and connected), and install the ` music-page` branch of skyra.pw for the dashboard part.

src/lib/structures/music/Queue.ts
```diff
@@ -135,6 +136,18 @@ export class Queue extends Array<Song> {
...
 
+	public toJSON() {
+		return {
+			length: this.length,
+			voiceChannel: this.voiceChannel,
+			playing: this.playing,
+			song: this.song,
+			position: this.position,
+			status: this.status,
+			queue: [...this]
```
PyroTechniac, line 147: Won't this cause circulars unless you call toJSON on each song?

kyranet: Nope.

src/lib/websocket/index.ts
```diff
@@ -0,0 +1,293 @@
+import WebSocket, { Server, Data } from 'ws';
```
favna: I'd prefer if this file be split into multiple smaller files so it's not 294 lines of blurb. Move the already exported interfaces and enums to `constants.ts` or `interfaces.ts`, the class `WebsocketHandler` to `Handler.ts` and the class `DashboardWebsocketUser` to `DashboardUser.ts`.

src/lib/websocket/index.ts
```diff
@@ -0,0 +1,293 @@
...
+import * as http from 'http';
```
favna:
```suggestion
import http from 'http';
```

This can be default-imported thanks to the `allowSyntheticDefaultImports` TS option that is active.

src/lib/websocket/index.ts
```diff
@@ -0,0 +1,293 @@
...
+import { SkyraClient } from '../SkyraClient';
+import Collection from '@discordjs/collection';
+import { enumerable } from '../util/util';
+import { Util } from 'klasa-dashboard-hooks';
+import { CLIENT_SECRET } from '../../../config';
+import { KlasaUser } from 'klasa';
+import { Events } from '../types/Enums';
+
+
+// TODO - should we timeout connections? disconnect after X period?
```
favna:
```suggestion
// TODO (Magnaboy): Should we timeout connections? disconnect after X period?
```

src/lib/websocket/index.ts
```diff
@@ -0,0 +1,293 @@
...
+// TODO - after the connection of a DashboardWebsocketUser is closed, is that class instance GC'd? should be
```
favna:
```suggestion
// TODO (Magnaboy): After the connection of a DashboardWebsocketUser is closed, is that class instance GC'd? should be
```

Conventional way of writing TODO messages

favna (commented):
Got a few remarks + what I send in Discord:

> There is one thing that I noticed in the music-api PR that could be done differently but I want to know what you guys think about it. This is the diff:
> So instead of  doing an `as SkyraClient` in `Queue.ts`, instead the augment is made in module augmentation in `SkyraGuild.ts`.  
>
> 
> 20-33-57_19-12-2019_Code

src/lib/websocket/index.ts
```diff
@@ -0,0 +1,48 @@
...
+import http from 'http';
+import { SkyraClient } from '../SkyraClient';
+import Collection from '@discordjs/collection';
+import { enumerable } from '../util/util';
+
+import WebsocketUser from './WebsocketUser';
+import { CloseCodes } from './types';
+
+// TODO - should we timeout connections? disconnect after X period?
+// TODO - after the connection of a WebsocketUser is closed, is that class instance GC'd? should be
```
Favna:
```suggestion
// TODO (Magnaboy): after the connection of a WebsocketUser is closed, is that class instance GC'd? should be
```

favna (commented):
\<snip>

src/lib/websocket/index.ts
```diff
@@ -0,0 +1,48 @@
...
+import http from 'http';
+import { SkyraClient } from '../SkyraClient';
+import Collection from '@discordjs/collection';
+import { enumerable } from '../util/util';
+
+import WebsocketUser from './WebsocketUser';
+import { CloseCodes } from './types';
+
+// TODO (Magnaboy): Should we timeout connections? disconnect after X period?
+// TODO - after the connection of a WebsocketUser is closed, is that class instance GC'd? should be
```
favna:
```suggestion
// TODO (Magnaboy): After the connection of a WebsocketUser is closed, is that class instance GC'd? should be
```

kyranet (approved):

PR #671, opened 2019-12-18
feat: new cache provider

PyroTechniac:


src/lib/queries/json.ts
```diff
@@ -14,7 +14,7 @@ export class JsonCommonQuery implements CommonQuery {
 
 	private client: Client;
 	private get provider() {
-		return this.client.providers.get('json') as JsonProvider;
+		return (this.client.providers.get('json') ?? this.client.providers.get('cache')) as JsonProvider;
```
PyroTechniac: This breaks when the cache provider is used over the JSON provider, as the JSON provider will still be loaded, which means it can't be `null` or `undefined`

kyranet: This is what I meant:
```suggestion
		return this.client.providers.default as JsonProvider;
```

A safer alternative:
```suggestion
		return this.client.providers.get(this.client.options.providers.default === 'json' ? 'json' : 'cache') as JsonProvider;
```

kyranet (dismissed):

src/lib/queries/json.ts
PyroTechniac: Close, it would be `this.client.providers.default.name`, if you can edit this I can commit it easily.

PyroTechniac:
This is also based on the rewrite for Settings Gateway, so this provider may not work completely until #646 is done.

favna (dismissed):

PR #674, opened 2019-12-22
feat(Games): Higher/Lower

kyranet:


cfanoulis (dismissed):
Looks better to me.

src/commands/Games/higherLower.ts
```diff
@@ -0,0 +1,267 @@
...
+import { Time } from '../../lib/util/constants';
+import { LLRCData, LongLivingReactionCollector } from '../../lib/util/LongLivingReactionCollector';
+import { getColor, resolveEmoji } from '../../lib/util/util';
+
+const enum HigherLowerReactions {
+	Higher = 'a:sarrow_up:658450971655012363',
+	Lower = 'a:sarrow_down:658452292558913536',
+	Cancel = ':redCross:637706251257511973',
+	Ok = ':greenTick:637706251253317669',
+	Cashout = '%F0%9F%92%B0' // 💰
```
favna, line 15: Why not just the emoji as a string?

src/commands/Games/higherLower.ts
```diff
@@ -0,0 +1,267 @@
+import { MessageEmbed } from 'discord.js';
+import { CommandStore, KlasaMessage } from 'klasa';
+import { SkyraCommand } from '../../lib/structures/SkyraCommand';
+import { Events } from '../../lib/types/Enums';
+import { UserSettings } from '../../lib/types/settings/UserSettings';
+import { Time } from '../../lib/util/constants';
+import { LLRCData, LongLivingReactionCollector } from '../../lib/util/LongLivingReactionCollector';
+import { getColor, resolveEmoji } from '../../lib/util/util';
+
+const enum HigherLowerReactions {
```
favna, line 10: shouldn't this be SCREAM_CASED?

src/commands/Games/higherLower.ts
```diff
@@ -0,0 +1,267 @@
...
+		}
+
+		const reaction = await this.awaitReactionAction(game);
+		await game.response.reactions.removeAll();
+
+		const action = reaction === null
+			? EndingAction.Timeout
+			: reaction === HigherLowerReactions.Ok
+				? EndingAction.Play
+				: EndingAction.Stop;
```
favna: nested ternaries are kind of ew and harder to read. If this can be refactored that would be nice.

> action is, if reaction strictly equals null, EndingAction.Timeout unless that is falsey then if reaction strictly equals HigherLowerReactions.Ok is truthy it is EndingAction.play else if false then it's EndingAction.stop

Just to put it to words to clarify why it's vague.

src/commands/Games/higherLower.ts
```diff
@@ -0,0 +1,267 @@
...
+
+		// Decide whether we timeout, stop, or continue
+		switch (action) {
+			case EndingAction.Timeout:
+				await game.response.edit(language.tget('COMMAND_HIGHERLOWER_TIMEOUT'), { embed: null });
```
favna: The lack of a `break;` here is on purpose?

src/commands/Games/higherLower.ts
```diff
@@ -0,0 +1,267 @@
...
+		await game.response.reactions.removeAll();
+
+		const action = reaction === null
+			? EndingAction.Timeout
+			: reaction === HigherLowerReactions.Ok
+				? EndingAction.Play
+				: EndingAction.Stop;
+
+		// Decide whether we timeout, stop, or continue
+		switch (action) {
```
favna: No default case is on purpose?

src/commands/Games/higherLower.ts
```diff
@@ -0,0 +1,267 @@
...
+
+			if (game.running) game.turn++;
+		}
+	}
+
+	private async win(game: HigherLowerGameData, message: KlasaMessage) {
+		const { language } = message;
+		await game.response.reactions.removeAll();
+
+		// TODO(Quantum): Implement Winning event for InfluxDB
```
favna:
```suggestion
		// TODO (Quantum): Implement Winning event for InfluxDB
```

Space for consistency with other TODO's

src/commands/Games/higherLower.ts
```diff
@@ -0,0 +1,267 @@
...
+
+		// There's a 0.1% chance that a user would lose now only the wager, but also what they would've won. Muahaha
+		if ((Math.random()) < 0.001) {
+			losses += this.calculateWinnings(wager, turn - 1);
+			await message.author.settings.decrease(UserSettings.Money, losses);
+		}
+
+		const { TITLE, DESCRIPTION, FOOTER } = message.language.tget('COMMAND_HIGHERLOWER_LOSE');
+		await game.response.edit(null, new MessageEmbed()
+			.setColor(getColor(message))
```
favna: `getColor(message)` should be moved to a private property so it doesn't have to be resolved 4x. Though I suppose the message is not accessible in the constructor, is it? In that case I suppose set it at the start of the run() and reference it in the other private methods where it's being used.

src/lib/util/util.ts
```diff
@@ -9,15 +9,14 @@ import nodeFetch, { RequestInit, Response } from 'node-fetch';
...
-import { APIEmojiData, APIUserData } from '../types/DiscordAPI';
+import { APIUserData } from '../types/DiscordAPI';
 import { Events } from '../types/Enums';
 import { GuildSettings } from '../types/settings/GuildSettings';
 import { UserSettings } from '../types/settings/UserSettings';
 import { UserTag } from './Cache/UserTags';
 import { BrandingColors, Time } from './constants';
 import { REGEX_UNICODE_BOXNM, REGEX_UNICODE_EMOJI } from './External/rUnicodeEmoji';
 import { LeaderboardUser } from './Leaderboard';
-import { LLRCDataEmoji } from './LongLivingReactionCollector';
```
favna, line 20: is `LLRCDataEmoji` still used elsewhere or can it be purged entirely from the `LongLivingReactionCollector` file?

favna (changes requested):

src/commands/Games/higherLower.ts
gc: Same reason we use enums for most things, although for here, isnt there a central Emoji  enum instead of having the strings in this file? For example, if we want to use :greenTick: or Cashout in another place, we'll have duplication of the emoji string

favna: I meant `Cashout = '💰'`, which would be perfectly valid, save a comment and achieve the exact same.

kyranet: No, it wouldn't.

All emojis are validated using the `encodeURIComponent`:

https://github.com/kyranet/Skyra/blob/1c9accc2ee56e1a8467f5240cf8d82b2fabaf179/src/lib/util/util.ts#L110-L121

So when it gets `'💰'`, it'll be turned into `'%F0%9F%92%B0'`, which would later mismatch.

src/commands/Games/higherLower.ts
kyranet: We use C# style here, not C++.

C# uses PascalCase, C++ uses SCREAMING_UPPER_CASE, which I dislike.

src/commands/Games/higherLower.ts
kyranet: > Action is, if reaction is strictly null, a timeout, otherwise if it was ok, then it'll continue the game, else it'll stop.

This works because LLRC validates all data before it's being emitted, making `reaction` one of `null | HigherLowerReactions.Ok | HigherLowerReactions.Cancel`.

Sure we can remove the ternary, I can move it to the switch, will be weird though.

src/commands/Games/higherLower.ts
kyranet: That's from @cfanoulis's code, you didn't remark it in his PR, but... half yes, half no.

It's supposed to call the end method below, which is right as it terminates the LLRC and everything, but contains one flaw, it re-edits this message.

I guess I'll have to do a couple of methods more so Timeout and Cancel share the common code while acting on its own.

src/commands/Games/higherLower.ts
kyranet: Yeah, it's unreachable, but I can make it throw as 100% of the times it'll be a bug.

src/commands/Games/higherLower.ts
kyranet: Can be moved to the game context, no problem.

src/commands/Games/higherLower.ts
```diff
@@ -0,0 +1,267 @@
...
+			running: true,
+			turn: 1,
+			number: this.random(0),
+			wager,
+			emojis: this.kFirstReactionArray,
+			callback: null
+		};
+
+		while (game.running) {
+			await game.response.reactions.removeAll();
```
vladfrangu: Is this really efficient? You're removing all reactions only to re-react... That's 1 DELETE, and then `n` PUTs for each emoji... versus at most one DELETE for the user reaction on an emoji...

EDIT: Either way you need Manage Messages sooo...

src/commands/Games/higherLower.ts
```diff
@@ -0,0 +1,267 @@
...
+		// TODO(Quantum): Implement Winning event for InfluxDB
+		const { TITLE, DESCRIPTION, FOOTER } = message.language.tget('COMMAND_HIGHERLOWER_WIN');
+		await game.response.edit(null, new MessageEmbed()
+			.setColor(getColor(message))
+			.setTitle(TITLE)
+			.setDescription(DESCRIPTION(this.calculateWinnings(game.wager, game.turn), game.number))
+			.setFooter(FOOTER));
+
+		// Ask the user whether they want to continue or cashout
+		game.emojis = this.kWinReactionArray;
```
vladfrangu: There's a LOT of code dupe here... This same or similar code is used above where I left my first review

This should probably be abstracted to one main function, no?

vladfrangu (changes requested):

src/lib/util/util.ts
kyranet: Yup:

https://github.com/kyranet/Skyra/blob/1c9accc2ee56e1a8467f5240cf8d82b2fabaf179/src/lib/util/Games/ConnectFour/PlayerHuman.ts#L51

src/commands/Games/higherLower.ts
```diff
@@ -0,0 +1,260 @@
...
+			// Send the embed
+			const { TITLE, DESCRIPTION, FOOTER } = message.language.tget('COMMAND_HIGHERLOWER_EMBED');
+			await game.response.edit(null, new MessageEmbed()
+				.setColor(game.color)
+				.setTitle(TITLE(game.turn))
+				.setDescription(DESCRIPTION(game.number))
+				.setFooter(FOOTER));
+
+			// Add the options
+			const emojis = game.turn > 1 ? this.kReactionArray : this.kFirstReactionArray;
```
cfanoulis, line 87: Shouldn't this use the game.emojis

kyranet: No, that's done this way to improve readability.

cfanoulis: ok /resolve

favna (approved):

PR #693, opened 2019-12-26
feat: queue now uses richdisplay

cfanoulis:
test img haha


PyroTechniac (changes requested):
2 things
1. Localization
2. This should also use the UserRichDislplay, not the built in RichDisplay

PR #720, opened 2020-01-03
feat(ws): Make WSU not be keyed by IP

kyranet:


kyranet:
A hard limit of 5 connections per IP address has been set in NGINX, this is deployed in production to avoid DDoS with this patch:

Docs: http://nginx.org/en/docs/http/ngx_http_limit_conn_module.html#limit_conn

kyranet:
Turns out, this causes issues, I'll revert the nginx configuration and do this in another way later.

PR #723, opened 2020-01-04
build: utilize module aliases

favna:
Have fun reviewing.

Marked as high priority because all other branches will need to be rebased after this


tsconfig.json
```diff
@@ -9,7 +9,16 @@
...
-		]
+		],
+		"baseUrl": ".",
+		"paths": {
+			"@root/*": ["./*"],
+			"@tasks/*": ["src/tasks/*"],
+			"@routes/*": ["src/routes/*"],
+			"@commands/*": ["src/commands/*"],
+			"@lib/*": ["src/lib/*"],
+			"@util/*": ["src/lib/util/*"]
```
kyranet: Do you know what bothers me? All are relative paths, but one starts with `./` and the other don't.

kyranet (changes requested):

src/Skyra.ts
```diff
@@ -1,6 +1,7 @@
+import { SkyraClient } from '@lib/SkyraClient';
+import { CLIENT_OPTIONS, TOKENS } from '@root/config';
+import 'module-alias/register';
```
vladfrangu: Shouldn't you register before importing with `@lib`?

vladfrangu (changes requested):

src/Skyra.ts
favna: Fixed in b461c4e, I noticed it at the same time

kyranet (approved):

PR #727, opened 2020-01-05
feat(starboard): Add user filter to the star command

### LINKED ISSUE #444: Feat(Starboard): Add user overload in the star command.
To check user's own stars.

kyranet:
Fixes #444


favna (commented):
Stars are guild specific so I suggest using membername over username as arg type

favna (approved):

PR #741, opened 2020-01-08
misc: Redesigned whois command

### LINKED ISSUE #737: request: restyle whois response to look better aesthetically
Feedback taken from #feedback (https://canary.discordapp.com/channels/254360814063058944/262740083977879552/664232241290805336):

> While this provides useful information, in my opinion it takes a quick glance to determine what is there

image
 
> **[23:08]CleverSource**:Like it's somewhat odd to have the data provided highlighted but then have the actual field of that data faded out, what I mean is:
> `Field           ::` **Data**
> **[23:08]CleverSource**:You quickly just notice the highlighted piece and then need to basically look over to determine what that data is.
> **[23:10]CleverSource**:I also feel like data is repeated unnecessarily, such as the user's snowflake being on the ID field and then the footer also has the user's snowflake.
> **[23:12]CleverSource**:and I feel like instead of the title being a link to the user's profile picture, it could instead be a mention to them, while the author can be their profile picture and username.
> **[23:21]CleverSource**:Here is an example of what it could look like - it has all the same information - pretend that the `@kyra` is a real mention and that the roles were mentioned - obviously not legitimately mentioned, just doing `<@&role_id>`.
> 
> On the "joined" and "account created" section, it would also be useful to have more than just a date, it could say something like "2016/12/02 - 1 year, 2 months, and 5 days ago" so you know how long ago the account joined or was created in a more human readable format.
 
image
 
 
> **[23:23]CleverSource**:I personally think it would look nicer and be easier to just quickly get some information out of it compared to how you currently have some of the command organized "aesthetic" wise.
> **[23:23]CleverSource**:and like I said before, I just think the stats command could use an embed, maybe you could possibly go through and just make things like a bit neater I guess, but none-the-less I hope this feedback helps you.
> **[23:23]CleverSource**:Hopefully that helps. :blobthumbsup:

kyranet:
Fixes #737

### DIFF
--- a/src/commands/Tools/avatar.ts
+++ b/src/commands/Tools/avatar.ts
@@ -1,7 +1,7 @@
 import { SkyraCommand } from '@lib/structures/SkyraCommand';
-import { getColor } from '@utils/util';
-import { MessageEmbed } from 'discord.js';
-import { CommandStore, KlasaMessage, KlasaUser } from 'klasa';
+import { getColor, getDisplayAvatar } from '@utils/util';
+import { MessageEmbed, GuildMember } from 'discord.js';
+import { CommandStore, KlasaMessage } from 'klasa';
 
 export default class extends SkyraCommand {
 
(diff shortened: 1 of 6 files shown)


favna:
Webp/PNG stuff from avatar should be in here too. Can't see avatar on mobile.

favna:
I'll do the above btw. ~~Going to extract it to a utility getAvatar() just like getColor()~~ There was already such a helper function

PR #742, opened 2020-01-08
feat: google search&image commands

favna:
This will require some API works on Google Custom Search engine (https://cse.google.com/cse/all) and Google Cloud Platform (https://console.developers.google.com/apis/library/customsearch.googleapis.com + api key stuff in IAM). As suggested in DMs on Discord I can help with that @kyranet. 

Two ~~open~~ closed question: 

1. Can I rename `search` command to `dsearch`, then give `gsearch.ts` an alias of `search` so google will be default, but DuckDuckGo will remain as alternative. 
  **A: Yes change**
2. Right now the URL for the image command  goes to the "context", for example if the image is a youtube thumbnail it will open the youtube video, not the image. I personally prefer this but I can also see some good of having the URL go to the actual image.
  **A: Keep way it's coded now**


favna:
Assigned WIP label until 2 open questions are answered

kyranet (approved):

PR #753, opened 2020-01-13
misc: Added colors enum to constants

kyranet:
Though it also changes some colors, for example, message delete logs will be red instead of orange, as requested by user feedback.


favna (commented):
Slotmachine also has some color constants, maybe replace those too?

other than that, LGTM.

kyranet:
That'll be a little complicated, canvas doesn't take numbers, so I'd have to turn them into formatted strings.

gc (approved):

favna (approved):

PR #759, opened 2020-01-13
refactor(CacheTags): Use RequestHandler to avoid fetching twice

kyranet:


src/lib/util/Cache/MemberTags.ts
```diff
@@ -118,6 +130,14 @@ export class MemberTags extends Collection<string, MemberTag> {
 		return highest;
 	}
 
+	private requestHandlerGet(id: string) {
+		return this.guild.members.fetch(id);
+	}
+
+	private requestHandlerGetAll(ids: readonly string[]) {
+		return Promise.all(ids.map(id => this.guild.members.fetch(id)));
```
vladfrangu, line 143: After https://github.com/discordjs/discord.js/pull/3562

```suggestion
		return this.guild.members.fetch({ user: ids });
```

vladfrangu (dismissed):

src/lib/util/Cache/MemberTags.ts
kyranet: I added a note in 94e68e4, thanks!

src/lib/util/Cache/MemberTags.ts
```diff
@@ -48,16 +54,22 @@ export class MemberTags extends Collection<string, MemberTag> {
 	public async fetch(): Promise<this>;
 	public async fetch(id?: string): Promise<MemberTag | null | this> {
 		if (typeof id === 'undefined') {
-			const members = await this.guild.members.fetch();
-			for (const member of members.values()) this.create(member);
+			if (this.kFetchAllPromise === null) {
+				this.kFetchAllPromise = this.guild.members.fetch().then(members => {
+					for (const member of members.values()) this.create(member);
+					this.kFetchAllPromise = null;
```
PyroTechniac: This should be moved to a `finally` function in case the promise throws, as to prevent an ever hanging promise.

PyroTechniac (changes requested):

src/lib/util/Cache/MemberTags.ts
```diff
@@ -48,16 +54,22 @@ export class MemberTags extends Collection<string, MemberTag> {
...
+			await this.kFetchAllPromise;
 			return this;
 		}
 
 		const existing = this.get(id);
 		if (typeof existing !== 'undefined') return existing;
 
 		try {
-			const member = await this.guild.members.fetch(id);
+			const member = await this.kRequestHandler.push(id);
```
PyroTechniac: Can't this block also be changed to use the new util function to ignore expected DAPI errors?

src/lib/util/Cache/UserTags.ts
```diff
@@ -56,17 +61,22 @@ export class UserTags extends Collection<string, UserTag> {
...
+		const userTag = await this.fetch(id);
+		return [id, userTag] as const;
 	}
 
 	public static get [Symbol.species](): CollectionConstructor {
 		return Collection as unknown as CollectionConstructor;
 	}
 
+	private requestHandlerGet(id: string) {
+		return this.client.users.fetch(id);
```
PyroTechniac, line 73: Same as above

PyroTechniac (changes requested):

src/lib/util/Cache/MemberTags.ts
kyranet: Resolved in f011b56.

src/lib/util/Cache/MemberTags.ts
kyranet: Resolved in 05541f9.

src/lib/util/Cache/UserTags.ts
kyranet: Uh... the original code doesn't catch any errors at all, and that goes out of this PR's scope.

kyranet: Also, RequestHandler's functions require objects with a property of `id`, when an error matches any of the codes given to `resolveOnErrorCodes`, it returns `null`, which doesn't match the trait `IdKeyed` from `@klasa/request-handler`.

vladfrangu (approved):

PR #767, opened 2020-01-15
refactor: de-embed slots and wof

favna:


kyranet (approved):
image

Landing as fast as the Boeing 767!

PR #782, opened 2020-01-19
Insert Elasticsearch into master

Quantumlyy:


src/lib/SkyraClient.ts
```diff
@@ -5,6 +5,8 @@ import { Client as VezaClient } from 'veza';
 import { Webhook } from 'discord.js';
 import { FSWatcher } from 'chokidar';
 import { DashboardClient } from 'klasa-dashboard-hooks';
+import { InfluxDB } from 'influx';
```
favna, line 8: This import can go, no? Are we still going to use Influx for anything else?

Rexogamer:
why the pr close?

kyranet (approved):

PR #812, opened 2020-02-08
test: unit testing for lib and such

### LINKED ISSUE #546: bug: missing i18n keys
**Describe the bug**
Several missing i18n keys, specifically from the SettingsMenu
**To Reproduce**
N/A

**Expected behavior**
N/A

**Screenshots**

**Additional context**
None

favna:
This comes with some pretty major refactors, so this PR will get a SEM: Major label. However these refactors are required to have a stable, easy to work with testing environment that is also scalable to the varied developers that work on Skyra. So without further ado the changes made:

1. Ensured that en-US and es-ES have all keys between them and all in the same sorting order. This was possible to track down thanks to the unit test!
1. Implemented TypeScript composite builds and project references. I tried to avoid it but it turned out to be a must-have requirement because otherwise ts-jest wouldn't pick up on things like module augmentation.
    - Updated all the scripts in package.json
    - Updated the tsc workflow for this. I can atest this works because I use it myself (and iCrawl and I had to go through various iterations because what even is testing)
    - Enabbled outputting declaration files from source. This is a requirement that comes with composite builds.
2. As side effect of that, `config.ts` and `config.example.ts` now reside in `src`. This is because they need to be in the same directory as the composite "src" project
3. Added the `__tests__` and `__mocks__` directories, which are default directory names for Jest.
4. Updated linting and building workflows to remove some old params for the respective actions.
5. Setup Jest and covered language files to the most extend that can be done.

resolves #546

Note: Merging this will likely cause merge conflicts to most other PR's, if not all!


src/languages/es-ES.ts
```diff
@@ -2519,12 +2547,12 @@ export default class extends Language {
...
-			HOW_MUCH: /^¿?cu[áa]nto/i,
-			HOW_MANY: /^¿?cu[áa]nto/i,
-			WHY: /^¿?por qu[ée]/i,
-			WHO: /^¿?qui[ée]n/i
+			WHEN: 'cuando',
+			WHAT: 'qué',
+			HOW_MUCH: 'cuanto',
+			HOW_MANY: 'cuantos',
+			WHY: 'por qué',
+			WHO: 'quien'
```
kyranet: ... why? Those changes are wrong

src/lib/queries/postgres.ts
```diff
@@ -64,6 +64,7 @@ export class PostgresCommonQuery implements CommonQuery {
 		`, [guildID, messageID]);
 	}
 
+	// eslint-disable-next-line @typescript-eslint/require-await
 	public async deleteMemberSettings(guildID: string, userID: string) {
```
kyranet: Just remove the `async`

src/lib/queries/postgres.ts
```diff
@@ -137,6 +138,7 @@ export class PostgresCommonQuery implements CommonQuery {
 		`);
 	}
 
+	// eslint-disable-next-line @typescript-eslint/require-await
 	public async purgeTwitchStreamGuildSubscriptions(guildID: string) {
```
kyranet: Ditto

__tests__/i18n.test.ts
```diff
@@ -0,0 +1,34 @@
...
+ * @param prefix Prefix to apply to each entry
+ */
+const i18nReducer = (obj: any, prefix = ''): any => Object.keys(obj).reduce((res: string | any[], el: string) => {
+	if (Array.isArray(obj[el])) {
+		return res;
+	} else if (typeof obj[el] === 'object' && obj[el] !== null) {
+		return [...res, ...i18nReducer(obj[el], `${prefix}${el}.`)];
+	}
+	return [...res, prefix + el];
+}, []);
```
kyranet: This looks a lot like Klasa's `objectToTuples`, except far more compute expensive (`@klasa/utils`'s `objectToTuples` iterates only once, this one does Object.keys, reduces, and concatenates, which are 3 iterations).

Maybe adapt the klasa code? https://github.com/dirigeants/utils/blob/master/src/lib/objectToTuples.ts

kyranet (changes requested):
I don't understand the need of those mocks, we already have the classes in `src`... And `@klasa/settings-gateway` exporting `Client` is a bug I'll have to fix, so don't rely on it.

__tests__/i18n.test.ts
```diff
@@ -0,0 +1,34 @@
...
+ * @param obj Object to deep reduce to keys
+ * @param prefix Prefix to apply to each entry
+ */
+const i18nReducer = (obj: any, prefix = ''): any => Object.keys(obj).reduce((res: string | any[], el: string) => {
+	if (Array.isArray(obj[el])) {
+		return res;
+	} else if (typeof obj[el] === 'object' && obj[el] !== null) {
+		return [...res, ...i18nReducer(obj[el], `${prefix}${el}.`)];
+	}
+	return [...res, prefix + el];
```
Quantumlyy:
```suggestion
	return [...res, `${prefix}${el}`];
```

package.json
```diff
@@ -84,11 +85,13 @@
...
 		"eslint-config-bamboo": "^3.0.0",
+		"jest": "^25.1.0",
+		"ts-jest": "^25.2.0",
 		"typescript": "^3.7.5"
 	},
 	"_moduleAliases": {
-		"@utils": "dist/src/lib/util",
-		"@lib": "dist/src/lib",
+		"@utils": "dist/lib/util",
+		"@lib": "dist/lib",
```
Quantumlyy, line 94: Why was this needed? It seemed correct the way it was previously. Mind explaining so it is visible if any other reviews request it?

Quantumlyy (changes requested):

package.json
kyranet: Because TSC, when suggesting those paths, it doesn't do in order of length, but in key insertion order.

By placing `@utils` before `@lib`, it'll suggest `@utils/util.ts` before `@lib/utils/util.ts`, and automatic import will take the first one.

src/languages/es-ES.ts
favna: They now match the same keys for English where they were literal strings rather than Regex expressions as well. Either one of them had to have been wrong. I assumed it was the Spanish.

package.json
favna: Unless you meant that it was `dist/src/lib` before, in which case it's because with this change there is no `src` directory in `dist` anymore as the `rootDir` TSC config option has been set.

src/languages/es-ES.ts
kyranet: That object supports either strings or RegExp, please revert that change

__tests__/i18n.test.ts
favna: Adopting Kyra's feedback instead about the util function. Resolving this one.

package.json
Quantumlyy: That is what it was directed it. So I ask why was dist/src/ moved to just dist/

package.json
```diff
@@ -2,23 +2,23 @@
 	"name": "skyra",
 	"version": "5.2.3",
 	"description": "Multipurpose Discord Bot built on Klasa",
-	"main": "dist/src/Skyra.js",
+	"main": "dist/Skyra.js",
```
Rexogamer, line 5: Is there such a file? Thought everything was in TS.

.vscode/launch.json
```diff
@@ -1,35 +1,35 @@
...
-    ]
-}
+	"version": "0.2.0",
+	"configurations": [
+		{
+			"type": "node",
+			"request": "launch",
+			"name": "Main - Windows",
+			"args": [
+				"${workspaceFolder}\\dist\\src\\Skyra.js"
```
Rexogamer: Same thing here

package.json
kyranet: Yeah, after compiling Skyra from TS to JS, `dist/Skyra.js` exists

.vscode/launch.json
kyranet:
```suggestion
				"${workspaceFolder}\\dist\\Skyra.js"
```

favna: gdi I searched for `src/Skyra`, forgetting about the `src\\Skyra` lol. Thanks for catching it guys.

__tests__/i18n.test.ts
```diff
@@ -0,0 +1,37 @@
...
+import Spanish from '@root/languages/es-ES';
+import { client } from '@testlib/MockInstances';
+
+/**
+ * Convert an object to a tuple.
+ * @description Adopted from `@klasa/utils`'s `objecToTuples`, but without including values which will never match between languages
+ * @param value The object to convert
+ * @param prefix The prefix for the key
+ */
+export default function objectToTuples<T = Record<string, unknown>>(original: T, prefix = ''): string[] {
```
favna:
```suggestion
function objectToTuples<T = Record<string, unknown>>(original: T, prefix = ''): string[] {
```

Woops, that was a copypasta mistake

vladfrangu (dismissed):

vladfrangu (dismissed):

.github/CONTRIBUTING.md
```diff
@@ -30,3 +30,6 @@ There are a number of guidelines considered when reviewing Pull Requests to be m
 - Everything should follow [OOP paradigms](https://en.wikipedia.org/wiki/Object-oriented_programming) and generally rely on behaviour over state where possible. This generally helps methods be predictable, keeps the codebase simple and understandable, reduces code duplication through abstraction, and leads to efficiency and therefore scalability.
 - Everything should follow our ESLint rules as closely as possible, and should pass lint tests even if you must disable a rule for a single line.
 - Everything should follow [Discord Bot Best Practices](https://github.com/meew0/discord-bot-best-practices)
+
+
```
kyranet: 🧹
```suggestion
```

.github/CONTRIBUTING.md
```diff
@@ -30,3 +30,6 @@ There are a number of guidelines considered when reviewing Pull Requests to be m
...
+[`config.example.ts`]:  /src/config.example.ts
```
kyranet:
```suggestion
[`config.example.ts`]: /src/config.example.ts
```

kyranet (changes requested):

src/tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+			"@lib/*": ["lib/*"],
+			"@root/*": ["*"]
+		},
+		"composite": true,
+		"declaration": true
+	},
+	"include": [
+		"./**/*"
+	]
+}
```
cfanoulis: pls add a finishing eol

tsconfig.eslint.json
```diff
@@ -0,0 +1,10 @@
+{
+	"extends": "./tsconfig.json",
+	"include": [
+		"./src/**/*",
+		"./database/**/*",
+		"./__tests__/**/*",
+		"./__mocks__/**/*",
+		"jest.config.js"
+	]
+}
```
cfanoulis: pls add a finishing eol

cfanoulis (changes requested):
general comments

kyranet (approved):

PR #820, opened 2020-02-11
chore: cleanup some ts-ignore's

favna:


kyranet (approved):
Hacky, but that works!

PR #857, opened 2020-03-01
Fix embed permission error

ImmortalSnake:
Some commands like `stats`, `support` and `topinvites` did not have permission checks for `EMBED_LINKS`

### DIFF
--- a/src/commands/Announcement/announcement.ts
+++ b/src/commands/Announcement/announcement.ts
@@ -19,6 +19,7 @@ export default class extends SkyraCommand {
 			extendedHelp: language => language.tget('COMMAND_ANNOUNCEMENT_EXTENDED'),
 			permissionLevel: PermissionLevels.Administrator,
 			requiredGuildPermissions: ['MANAGE_ROLES'],
+			requiredPermissions: ['EMBED_LINKS'],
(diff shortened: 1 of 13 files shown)


src/commands/Announcement/announcement.ts
```diff
@@ -19,6 +19,7 @@ export default class extends SkyraCommand {
 			extendedHelp: language => language.tget('COMMAND_ANNOUNCEMENT_EXTENDED'),
 			permissionLevel: PermissionLevels.Administrator,
 			requiredGuildPermissions: ['MANAGE_ROLES'],
+			requiredPermissions: ['EMBED_LINKS'],
```
kyranet: This also requires two more permissions, one to add reactions, and another to delete them:

```suggestion
			requiredPermissions: ['ADD_REACTIONS', 'MANAGE_MESSAGES', 'EMBED_LINKS'],
```

kyranet (changes requested):

src/commands/Announcement/announcement.ts
ImmortalSnake: Alright then

kyranet (approved):
LGTM, thanks!

PR #866, opened 2020-03-08
src: Fix grammar

vladfrangu:


kyranet (approved):

favna:
LGTM but while checks are pending can you also double check Spanish one for this key?

PR #887, opened 2020-03-25
feat: refactor a lot in moderation commands

favna:
- 3 new moderation commands, `setnickname`, `addrole` and `removerole`.
- fixed `vunmute`, `vmute` and `voicekick`
- refactored the way args are parsed and types are done for moderation commands
- use `@ApplyOptions` for most moderation commands
- use `@ApplyOptions` in the command vscode snippet
- fix the inclusion of `.d.ts` files in gitignore


src/commands/Moderation/removeRole.ts
```diff
@@ -0,0 +1,29 @@
...
+	permissionLevel: PermissionLevels.Administrator,
+	usage: '<user:username> <role:rolename> [reason:...string]',
+	usageDelim: ' '
+})
+export default class extends SkyraCommand {
+
+	public async run(message: KlasaMessage, [target, role, reason]: [User, Role, string | null]) {
+		await checkModeratable(message, target, true, this.client);
+
+		return message.guild!.security.actions.addRole({
```
kyranet: Shouldn't this be removeRole?

src/commands/Moderation/setNickname.ts
```diff
@@ -0,0 +1,40 @@
...
+			moderator_id: message.author.id,
+			reason
+		}, nickname, getTargetDM(message, target));
+	}
+
+	public async init() {
+		this.createCustomResolver('nickname', (arg, possible, message) => {
+			if (!arg) return '';
+
+			return this.kStringArgtype.run(arg, possible, message);
```
kyranet:
```suggestion
			return arg ? this.kStringArgtype.run(arg, possible, message) : '';
```

src/lib/util/Security/ModerationActions.ts
```diff
@@ -447,6 +471,36 @@ export class ModerationActions {
 		return this.removeStickyRole(moderatorID, id, mutedRole);
 	}
 
+	private async nicknameUser(rawOptions: ModerationActionOptions, nickname: string) {
+		try {
+			const member = await this.guild.members.fetch(rawOptions.user_id);
+			return await member.setNickname(nickname);
```
kyranet: Please change this to do a single direct REST call to add a role, instead of fetching and then add the role.

src/lib/util/Security/ModerationActions.ts
```diff
@@ -447,6 +471,36 @@ export class ModerationActions {
...
+		} catch (error) {
+			if (error.code === APIErrors.UnknownMember) throw this.guild.language.tget('ACTION_REQUIRED_MEMBER');
+			throw error;
+		}
```
kyranet: There's an utility that takes a Promise and makes it throw on unmatching APIErrors codes, returning `null` on those that match the codes passed, you can use it instead of wrapping this on a try/catch 👍

src/commands/Moderation/addRole.ts
```diff
@@ -0,0 +1,29 @@
...
+import { Role, User } from 'discord.js';
+import { KlasaMessage } from 'klasa';
+
+@ApplyOptions<SkyraCommandOptions>({
+	cooldown: 10,
+	description: language => language.tget('COMMAND_ADDROLE_DESCRIPTION'),
+	extendedHelp: language => language.tget('COMMAND_ADDROLE_EXTENDED'),
+	requiredGuildPermissions: ['MANAGE_ROLES'],
+	permissionLevel: PermissionLevels.Administrator,
+	usage: '<user:username> <role:rolename> [reason:...string]',
```
kyranet: Any reason to not support duration? AddRoles can be reversed (removing the role)

src/commands/Moderation/removeRole.ts
```diff
@@ -0,0 +1,29 @@
...
+import { Role, User } from 'discord.js';
+import { KlasaMessage } from 'klasa';
+
+@ApplyOptions<SkyraCommandOptions>({
+	cooldown: 10,
+	description: language => language.tget('COMMAND_REMOVEROLE_DESCRIPTION'),
+	extendedHelp: language => language.tget('COMMAND_REMOVEROLE_EXTENDED'),
+	requiredGuildPermissions: ['MANAGE_ROLES'],
+	permissionLevel: PermissionLevels.Administrator,
+	usage: '<user:username> <role:rolename> [reason:...string]',
```
kyranet: Any reason to not support duration? RemoveRoles can be reversed (adding the role back)

kyranet (changes requested):
All 3 commands should have `requiredMember: true`, as you can't execute them on members that are not in the guild.

src/commands/Moderation/removeRole.ts
favna: uh yeah, woops

src/lib/util/Security/ModerationActions.ts
kyranet: This one, specifically:
https://github.com/skyra-project/skyra/blob/0a4ca2ddd9669b41cb5f67a387d59a076f49da32/src/lib/util/util.ts#L524-L531

favna:
Barring duration all requests are fixed so I'm de-drafting this. If you want duration we're going to have to look together because I cannot figure out for the life of me how to extend ModerationCommand.

src/lib/structures/ModerationCommand.ts
```diff
@@ -145,3 +120,38 @@ export abstract class ModerationCommand<T = unknown> extends SkyraCommand {
 	public abstract posthandle(message: KlasaMessage, targets: User[], reason: string | null, prehandled: T): unknown;
 
 }
+
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export async function checkModeratable<T>(message: KlasaMessage, target: User, requiredMember: boolean, client: KlasaClient, _prehandled?: T) {
```
kyranet: You should retrieve client from `message`. Also, I'd make this an static method on ModerationCommand, so this is namespaced.

src/lib/structures/ModerationCommand.ts
```diff
@@ -145,3 +120,38 @@ export abstract class ModerationCommand<T = unknown> extends SkyraCommand {
...
+	if (member) {
+		const targetHighestRolePosition = member.roles.highest.position;
+		if (targetHighestRolePosition >= message.guild!.me!.roles.highest.position) throw message.language.tget('COMMAND_ROLE_HIGHER_SKYRA');
+		if (targetHighestRolePosition >= message.member!.roles.highest.position) throw message.language.tget('COMMAND_ROLE_HIGHER');
+	}
+
+	return member;
+}
+
+export function getTargetDM(message: KlasaMessage, target: User): ModerationActionsSendOptions {
```
kyranet: Same here (namespaced method)

kyranet (changes requested):
Besides all of this, it'd still be nice if we had duration in addRole and removeRole.

src/commands/Moderation/addRole.ts
```diff
@@ -0,0 +1,29 @@
...
+	usageDelim: ' '
+})
+export default class extends SkyraCommand {
+
+	public async run(message: KlasaMessage, [target, role, reason]: [User, Role, string | null]) {
+		await ModerationCommand.checkModeratable(message, target, true, this.client);
```
kyranet:
```suggestion
		await ModerationCommand.checkModeratable(message, target, true);
```

src/commands/Moderation/removeRole.ts
```diff
@@ -0,0 +1,29 @@
...
+	usageDelim: ' '
+})
+export default class extends SkyraCommand {
+
+	public async run(message: KlasaMessage, [target, role, reason]: [User, Role, string | null]) {
+		await ModerationCommand.checkModeratable(message, target, true, this.client);
```
kyranet:
```suggestion
		await ModerationCommand.checkModeratable(message, target, true);
```

kyranet (changes requested):

kyranet (approved):

PR #891, opened 2020-03-29
refactor: run yarn install in update command

favna:


src/commands/System/Admin/update.ts
```diff
@@ -27,6 +28,13 @@ export default class extends SkyraCommand {
 		return message.channel.send(`${Emojis.GreenTick} Successfully compiled.`);
 	}
 
+	private async updateDependencies(message: KlasaMessage) {
+		const { stderr } = await util.exec('yarn install --frozen-lockfile')
+			.catch(error => ({ stdout: '', stderr: (error && error.message) || error || '' }));
```
kyranet:
```suggestion
			.catch(error => ({ stdout: '', stderr: error?.message || error || '' }));
```

kyranet (changes requested):

kyranet (approved):

PR #912, opened 2020-04-01
feat: add role change logs

favna:
This adds logging for role changes

IMPORTANT: This will require adding a column to the Guilds database on production. I actually don't know how you do this...?

### DIFF
--- a/src/events/rawGuildMemberUpdate.ts
+++ b/src/events/rawGuildMemberUpdate.ts
@@ -1,3 +1,4 @@
+import { Colors } from '@lib/types/constants/Constants';
 import { AuditLogResult, WSGuildMemberUpdate } from '@lib/types/DiscordAPI';
 import { Events } from '@lib/types/Enums';
 import { GuildSettings } from '@lib/types/settings/GuildSettings';
@@ -7,7 +8,6 @@ import { api } from '@utils/Models/Api';
 import { floatPromise, getDisplayAvatar } from '@utils/util';
 import { MessageEmbed } from 'discord.js';
 import { Event, EventStore, KlasaGuild } from 'klasa';
-import { Colors } from '@lib/types/constants/Constants';
 
 export default class extends Event {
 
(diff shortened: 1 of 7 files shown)


src/events/rawGuildMemberUpdate.ts
```diff
@@ -22,34 +23,65 @@ export default class extends Event {
...
 		floatPromise(this, this.handleRoleSets(guild, data));
 	}
 
-	private handleNicknameChange(guild: KlasaGuild, data: WSGuildMemberUpdate) {
-		// Get the current nickname, compare them both, if they are different, it changed
+	private handleMemberChange(guild: KlasaGuild, data: WSGuildMemberUpdate) {
+
+		// Look up which logs should be send in the guild database
+		let shouldLogNickname = guild.settings.get(GuildSettings.Events.MemberNicknameUpdate);
+		let shouldSendRoleLog = guild.settings.get(GuildSettings.Events.MemberRoleUpdate);
```
kyranet: Those should be moved as they're not used until a few lines below, and can end up unused as of line 50:

```javascript
if (typeof previous === 'undefined') return;
```

src/events/rawGuildMemberUpdate.ts
```diff
@@ -22,34 +23,65 @@ export default class extends Event {
...
+				: null;
+
+			// Set the Role change log
+			this.client.emit(Events.GuildMessageLog, MessageLogsEnum.Member, guild, () => new MessageEmbed()
+				.setColor(Colors.Yellow)
+				.setAuthor(`${data.user.username}#${data.user.discriminator} (${data.user.id})`, getDisplayAvatar(data.user.id, data.user))
+				.setDescription(guild.language.tget('EVENTS_ROLE_DIFFERENCE', previousRoleNames, nextRoleNames))
+				.setFooter(guild.language.tget('EVENTS_ROLE_UPDATE'))
+				.setTimestamp());
+		}
```
kyranet: This should not log all roles, but the ones that were added or removed.
I think Skyra III had this code, might be worth a check? I can check in around 3 hours if I don't forget.

kyranet (changes requested):

kyranet:
> This will require adding a column to the Guilds database on production. I actually don't know how you do this...?

SettingsGateway [handles it for us](https://github.com/dirigeants/settings-gateway/blob/69a67e5d0b85ff2abf9ed9b16c0df861abaa38ac/src/lib/gateway/GatewayStorage.ts#L70-L78)

src/events/rawGuildMemberUpdate.ts
kyranet: Found it, v2.1.1:

https://github.com/skyra-project/skyra/blob/7d5a418bedfc026fefb2062774e284d0e79a423a/events/guildMemberUpdate.js#L59-L69

Code could be tidied up, it's a little messy, to be expected from +2 years old code.

src/events/rawGuildMemberUpdate.ts
```diff
@@ -22,34 +23,67 @@ export default class extends Event {
...
 				.setTimestamp());
 		}
+
+		// Retrieve whether Role Logs should be send from Guild Settings
+		let shouldSendRoleLog = guild.settings.get(GuildSettings.Events.MemberRoleUpdate);
+
+		// If role arrays are identical then skip role log
+		if (arrayStrictEquals(previous.roles, next.roles)) shouldSendRoleLog = false;
+
+		if (shouldSendRoleLog) {
```
kyranet:
```suggestion
		// Retrieve whether or not role logs should be sent from Guild Settings and
		// whether or not the roles are the same.
		if (guild.settings.get(GuildSettings.Events.MemberRoleUpdate) && !arrayStrictEquals(previous.roles, next.roles)) {
```

src/events/rawGuildMemberUpdate.ts
```diff
@@ -22,34 +23,67 @@ export default class extends Event {
...
 
-		// TODO(kyranet): Role Change Logs
-		if (guild.settings.get(GuildSettings.Events.MemberNicknameUpdate)) {
+		// Retrieve whether Nickname Logs should be send from Guild Settings
+		let shouldLogNickname = guild.settings.get(GuildSettings.Events.MemberNicknameUpdate);
+
+		// If nicknames are identical then skip nickname log
+		if (previous.nickname === next.nickname) shouldLogNickname = false;
+
+		if (shouldLogNickname) {
```
kyranet:
```suggestion
		// Retrieve whether or not nickname logs should be sent from Guild Settings and
		// whether or not the nicknames are identical.
		if (guild.settings.get(GuildSettings.Events.MemberNicknameUpdate) && previous.nickname !== next.nickname) {
```

kyranet (changes requested):

kyranet (approved):

PR #929, opened 2020-04-22
fix: Tweak tags.list to send with an URD

kyranet:


src/commands/Tags/tag.ts
```diff
@@ -1,59 +1,31 @@
...
-		this.createCustomResolver('tagname', (arg, possible, message, [action]) => {
-			if (action === 'list') return undefined;
-			if (!arg) throw message.language.tget('RESOLVER_INVALID_STRING', possible.name);
-			if (arg.includes('`') || arg.includes('\u200B')) throw message.language.tget('COMMAND_TAG_NAME_NOTALLOWED');
-			if (arg.length > 50) throw message.language.tget('COMMAND_TAG_NAME_TOOLONG');
-			return arg.toLowerCase();
-		});
-	}
-
+	@requiresPermission(4, (message: KlasaMessage) => { throw message.language.tget('COMMAND_TAG_PERMISSIONLEVEL'); })
```
favna: Add 4 to the PermissionLevels enum (`src/lib/types/Enums.d.ts`) and reference it here and the other annotations.

favna (changes requested):
Just 1 thing

favna (approved):

PR #930, opened 2020-04-22
fix(music): fix the ws events

favna:
This is required for https://github.com/skyra-project/skyra.pw/pull/154

---

Previously music ws events were untyped and data being send was extremely inconsistent which made it hard to process on the frontend. This PR solves these issues

Note: I don't know when `.toString()` and `.toJSON()` are automatically called so possibly some of those calls can still be removed.

### DIFF
--- a/src/events/music/musicRemove.ts
+++ b/src/events/music/musicRemove.ts
@@ -14,7 +14,7 @@ export default class extends Event {
 		}
 
 		for (const subscription of manager.websocketUserIterator()) {
-			subscription.send({ action: OutgoingWebsocketAction.MusicRemove, data: { id: song.id } });
+			subscription.send({ action: OutgoingWebsocketAction.MusicRemove, data: { song: song.toJSON() } });
 		}
 	}
 
--- a/src/events/music/musicRemove.ts
+++ b/src/events/music/musicRemove.ts
@@ -14,7 +14,7 @@ export default class extends Event {
 		}
 
 		for (const subscription of manager.websocketUserIterator()) {
-			subscription.send({ action: OutgoingWebsocketAction.MusicRemove, data: { song: song.toJSON() } });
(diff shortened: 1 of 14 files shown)


src/lib/util/Models/ApiTransform.ts
```diff
@@ -271,3 +273,62 @@ export interface FlattenedMember {
...
+	author: string;
+	duration: number;
+	stream: boolean;
+	position: number;
+	title: string;
+	url: string;
+	skips: string[];
+}
+
+export function flattedMusicHandler(handler: MusicHandler): FlattenedMusicHandler {
```
kyranet: Shouldn't this be flatten?

src/events/music/musicRemove.ts
```diff
@@ -14,7 +14,7 @@ export default class extends Event {
 		}
 
 		for (const subscription of manager.websocketUserIterator()) {
-			subscription.send({ action: OutgoingWebsocketAction.MusicRemove, data: { id: song.id } });
+			subscription.send({ action: OutgoingWebsocketAction.MusicRemove, data: { queue: manager.queue.map(s => s.toJSON()) } });
```
vladfrangu, line 17: Hear me out: Why is musicRemove and musicSongFinish returning the full queue? Surely it isn't that hard on the frontend to `queue.filter(s => s.id != packet.data.id)`.. At this point, you're sending the same payload through a lot of events, and might as well unify it into a MusicSongUpdate generic one..

vladfrangu (changes requested):

src/events/music/musicRemove.ts
favna: I thought so as well but for some godforsaken reason the frontend completely loses track of what the previous queue was when receiving new data and I've tried literally everything under the sun to resolve it so I ended up with this.

Believe me, if it could be done I would not have made this change.

vladfrangu: Link me the code that's affected and let me take a quick peek at it 👀

kyranet (approved):

PR #935, opened 2020-04-25
fix(urd): dont add page numbers when lacking permissions

favna:
I noticed when Skyra has no permission to add reactions that it still adds the page number count which is very misleading since users cannot do anything with the RichDisplay at that point. 

So instead of using Klasa's build in way to add a footer we should use this way to ensure that Skyra has the required permissions before setting the footer.

As for the changes on `anime`, `manga` and `urban`, they were just wrong before and never showing up.

### DIFF
--- a/src/lib/structures/UserRichDisplay.ts
+++ b/src/lib/structures/UserRichDisplay.ts
@@ -1,5 +1,5 @@
 import { Time } from '@utils/constants';
-import { Client, MessageEmbed, MessageReaction, TextChannel } from 'discord.js';
+import { Client, MessageEmbed, MessageReaction, Permissions, TextChannel } from 'discord.js';
 import { KlasaMessage, KlasaUser, ReactionHandler, RichDisplay, RichDisplayRunOptions, util } from 'klasa';
 
 export class UserRichDisplay extends RichDisplay {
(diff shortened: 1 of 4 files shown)


src/lib/structures/UserRichDisplay.ts
```diff
@@ -15,13 +20,23 @@ export class UserRichDisplay extends RichDisplay {
...
 		const handler = (await super.run(message, options))
 			.once('end', () => UserRichDisplay.handlers.delete(target));
 		UserRichDisplay.handlers.set(target, handler);
 
 		return handler;
 	}
 
+	private setAuthorizedFooter(client: Client, channel: TextChannel) {
+		const permissionsForClient = (channel.permissionsFor(client.user!)!);
+		if (permissionsForClient.has('ADD_REACTIONS') && permissionsForClient.has('MANAGE_MESSAGES')) {
```
kyranet: You could have:
```ts
private static readonly kPermissions = new Permissions([
	Permissions.FLAGS.ADD_REACTIONS,
	Permissions.FLAGS.MANAGE_MESSAGES
]);
```

And then in here:

```ts
const priviledged = channel.permissionsFor(client.user!)?.has(UserRichDisplay.kPermissions) ?? false;
if (priviledged) {
	// ...
}
```

kyranet (changes requested):

kyranet (approved):

PR #936, opened 2020-04-28
fix(content, eval): fixes for length

favna:
Previously if the content or evalled result was about 1990 characters
(6 backticks, the "md" and the enters) it would throw an error about
message length. To fix this I ported the uploading as file or hastebin
(leaving out logs) from Eval to Content. I also noticed Eval was
checking for a hard 2K characters which also caused issues for the same
reason so dropped it to 1980 as well. While I was working I also added
the last bit of translations for Eval since it was needed for Content
anyway.


src/languages/es-ES.ts
```diff
@@ -2268,6 +2268,9 @@ export default class extends Language {
 		}),
 		COMMAND_CONTENT_DESCRIPTION: 'Obtener el contenido sin formato de los mensajes.',
 		COMMAND_CONTENT_EXTENDED: builder.display('content', {}),
+		COMMAND_CONTENT_OUTPUT_FILE: 'Envió el resultado como un archivo.',
+		COMMAND_CONTENT_OUTPUT_HASTEBIN: url => `Envió el resultado a hastebin: ${url}`,
```
kyranet:
```suggestion
		COMMAND_CONTENT_OUTPUT_FILE: 'Enviado el resultado como un archivo.',
		COMMAND_CONTENT_OUTPUT_HASTEBIN: url => `Enviado el resultado a hastebin: ${url}`,
```

src/commands/Tools/content.ts
```diff
@@ -30,7 +32,69 @@ export default class extends SkyraCommand {
...
+		const _options = ['none', 'abort'];
+		if (message.channel.attachable) _options.push('file');
+		if (!options.hastebinUnavailable) _options.push('hastebin');
+		let _choice: { content: string };
+		do {
+			_choice = await message.prompt(message.language.tget('COMMAND_CONTENT_CHOOSE_OUTPUT', _options)).catch(() => ({ content: 'none' }));
+		}
+		while (!_options.concat('none', 'abort').includes(_choice.content));
+		options.sendAs = _choice.content.toLowerCase();
+	}
```
kyranet: Those should be moved to an utility file to avoid code duplication with the eval command.

kyranet (changes requested):

kyranet (approved):

PR #940, opened 2020-05-03
fix(overwatch): fix for no ratings

favna:
also fixed the player we mention in help, for some reason the previous one has no data available anymore so instead we now have a D.VA player, obviously automatically 10x better because who doesn't love korean bunny girl 😏

### DIFF
--- a/src/commands/GameIntegration/overwatch.ts
+++ b/src/commands/GameIntegration/overwatch.ts
@@ -54,7 +54,7 @@ export default class OverwatchCommand extends SkyraCommand {
 		platform: PlatformUnion
 	) {
 		const EMBED_DATA = message.language.tget('COMMMAND_OVERWATCH_EMBED_DATA');
-		const ratings = this.ratingsToMap(overwatchData.ratings, r => r.role, r => r);
+		const ratings = this.ratingsToMap(overwatchData.ratings ?? [], r => r.role, r => r);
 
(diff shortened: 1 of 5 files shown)


src/commands/GameIntegration/overwatch.ts
```diff
@@ -73,7 +73,7 @@ export default class OverwatchCommand extends SkyraCommand {
 				].join('\n'))
 				.addField(EMBED_DATA.RATINGS_TITLE, EMBED_DATA.RATINGS([
 					...ratings.values(),
-					{ role: 'average', level: overwatchData.rating }
+					{ role: 'average', level: overwatchData.rating === 0 ? EMBED_DATA.NO_AVERAGE : overwatchData.rating }
```
kyranet, line 76: Shouldn't this be checking the array's length?
```suggestion
					{ role: 'average', level: overwatchData.rating?.length ? EMBED_DATA.NO_AVERAGE : overwatchData.rating }
```

kyranet (changes requested):

src/commands/GameIntegration/overwatch.ts
favna: No, `rating` is a number (on the API data). `ratings` is an array (parsed from the API data)

kyranet (approved):

PR #941, opened 2020-05-03
fix(Remindme): removed NLP and use Klasa's argument parser instead

kyranet:


src/commands/Social/remindme.ts
```diff
@@ -68,46 +68,48 @@ export default class extends SkyraCommand {
...
 
-		let time: number;
-		let attempts = 0;
-		do {
-			const messages = await message.channel.awaitMessages(msg => msg.author === message.author, { time: 30000, max: 1 });
-			if (!messages.size) throw null;
-			time = new Duration(messages.first()!.content).offset;
-			attempts++;
-		} while (time < 60000 && attempts < 5);
+		this.createCustomResolver('idOrDuration', (arg, possible, message, [action]: Actions[]) => {
```
vladfrangu, line 94: The type for the action destructuring should be `[Actions]`

src/commands/Social/remindme.ts
```diff
@@ -68,46 +68,48 @@ export default class extends SkyraCommand {
...
+				case Actions.Create: {
+					if (!arg) throw message.language.tget('COMMAND_REMINDME_CREATE_NO_DURATION');
+					return this.client.arguments.get('timespan')!.run(arg, { ...possible, min: Time.Minute }, message);
+				}
+			}
+		});
 
-		if (!time || time < 60000) throw message.language.tget('COMMAND_REMINDME_SHORT_TIME');
-		return time;
+		this.createCustomResolver('description', (arg, possible, message, [action]: Actions[]) => {
```
vladfrangu, line 108: Ditto

vladfrangu (changes requested):

src/commands/Social/remindme.ts
kyranet: Not possible, it complains that it doesn't match `any[]`.

src/commands/Social/remindme.ts
kyranet: Ditto

favna (commented):
> can you also add a show action to reminders to show the full reminders text?

kyranet:
> > can you also add a show action to reminders to show the full reminders text?

Done in d453c77.

src/commands/Social/remindme.ts
```diff
@@ -68,46 +68,50 @@ export default class extends SkyraCommand {
...
-
-			if (indexOfTime !== -1) {
-				parsed.time = new Duration(input.slice(indexOfTime + 4)).offset;
+	public async init() {
+		this.createCustomResolver('action', (arg, _possible, message) => {
+			if (!arg) return Actions.List;
+
+			switch (arg.toLowerCase()) {
+				case 's':
+				case 'show':
```
favna:
```suggestion
```

favna (changes requested):

favna (approved):

PR #943, opened 2020-05-03
feat(CommandError): handle AbortError errors

kyranet:


favna (dismissed):

src/events/commandError.ts
```diff
@@ -4,6 +4,7 @@ import { inlineCodeblock } from '@utils/util';
 import { DiscordAPIError, HTTPError, MessageEmbed } from 'discord.js';
 import { Command, Event, KlasaMessage, util } from 'klasa';
 import { Colors } from '@lib/types/constants/Constants';
+import 'node-fetch';
```
favna: why this import btw?

kyranet: Remnants from tests (I wanted to check if it exposed AbortError, turns out they don't).

Nice catch though, thanks!

favna (approved):

PR #953, opened 2020-05-04
fix(Slowmode): resolved negative time bugs

kyranet:


src/commands/Moderation/Utilities/slowmode.ts
```diff
@@ -1,37 +1,37 @@
...
+		if (cooldown === 'reset' || cooldown === 'off' || cooldown < 0) cooldown = 0;
 		else if (cooldown >= MAXIMUM_TIME) throw message.language.get('COMMAND_SLOWMODE_TOO_LONG');
 		const channel = message.channel as TextChannel;
 		await channel.setRateLimitPerUser(cooldown);
 		return message.sendLocale('COMMAND_SLOWMODE_SET', [cooldown * 1000]);
 	}
 
+	public async init() {
+		this.createCustomResolver('cooldown', async (arg, possible, message) =>
+			(await this.client.arguments.get('timespan')!.run(arg, possible, message)) / 1000);
```
favna, line 34:
```suggestion
			(await kTimespanArg.run(arg, possible, message)) / 1000);
```
and create the private field: `private kTimespanArg = this.client.arguments.get('timespan')!`.
That way the look up in the store is cached...

favna (changes requested):

src/commands/Moderation/Utilities/slowmode.ts
kyranet: Actually, we should not define other pieces instances as a property, because of raceconditions and that we'd need to reload this command when we reload timespan.

favna (approved):

PR #963, opened 2020-05-04
feat(Webhooks): added webhook for Bots For Discord

kyranet:
I missed this when migrating the API from Ny-API to Skyra in 5.1.0, but fear not, it's back again!

### DIFF
--- a/src/routes/webhooks/b4d.ts
+++ b/src/routes/webhooks/b4d.ts
@@ -1,14 +1,12 @@
 import ApiRequest from '@lib/structures/api/ApiRequest';
 import ApiResponse from '@lib/structures/api/ApiResponse';
-import { Route, RouteStore } from 'klasa-dashboard-hooks';
+import { Route, RouteOptions } from 'klasa-dashboard-hooks';
 import { TOKENS } from '@root/config';
 import { UserSettings } from '@lib/types/settings/UserSettings';
+import { ApplyOptions } from '@skyra/decorators';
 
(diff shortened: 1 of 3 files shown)


favna (approved):

favna (changes requested):
On second thought, `@ApplyOptions` should work I think. Route extends Piece.

src/routes/webhooks/b4d.ts
```diff
@@ -0,0 +1,46 @@
...
+import ApiResponse from '@lib/structures/api/ApiResponse';
+import { Route, RouteStore } from 'klasa-dashboard-hooks';
+import { TOKENS } from '@root/config';
+import { UserSettings } from '@lib/types/settings/UserSettings';
+
+export default class extends Route {
+
+	public constructor(store: RouteStore, file: string[], directory: string) {
+		super(store, file, directory, { route: 'webhooks/b4d' });
+	}
```
favna:
```suggestion
import ApiRequest from '@lib/structures/api/ApiRequest';
import ApiResponse from '@lib/structures/api/ApiResponse';
import { ApplyOptions } from '@skyra/decorators';
import { Route, RouteOptions } from 'klasa-dashboard-hooks';
import { TOKENS } from '@root/config';
import { UserSettings } from '@lib/types/settings/UserSettings';

@ApplyOptions<RouteOptions>({ route: 'webhooks/b4d' })
export default class extends Route {

```

kyranet: Fixed in d992a73.

favna (approved):

PR #965, opened 2020-05-04
fix(Embeds): use dynamic png for images

### LINKED ISSUE #962: bug: avatars should be set with dynamic
**Describe the bug**

Avatars as webp's are not always displayed (in particular on iOS) so we should use `png` and `gif` instead. DJS has `.displayAvatarURL({ format: 'png', dynamic: true})` for this. We should also refactor out some of the manual resolving of avatar where possible (I know there is a util function)

**To Reproduce**
0. Be on iOS
1. Use Love Command
2. Avatar image is not displayed on embed

**Expected behavior**
Avatar image is displayed

kyranet:
Fixes #962

### DIFF
--- a/src/commands/Fun/love.ts
+++ b/src/commands/Fun/love.ts
@@ -38,7 +38,7 @@ export default class extends SkyraCommand {
 
 		return message.sendEmbed(new MessageEmbed()
 			.setColor(getColor(message))
-			.setAuthor('❤ Love Meter ❤', message.author.displayAvatarURL())
+			.setAuthor('❤ Love Meter ❤', message.author.displayAvatarURL({ size: 128, format: 'png', dynamic: true }))
 			.setThumbnail('https://twemoji.maxcdn.com/2/72x72/1f49e.png')
 			.setDescription([
 				`💗 **${user.tag}**`,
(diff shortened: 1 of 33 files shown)


favna:
We should alsp be able to refactor this from utils out by using dynamic I think, I remember it's being used at least once coming from a user and not a user tag.

```ts
const ROOT = 'https://cdn.discordapp.com';
export function getDisplayAvatar(id: string, user: UserTag | User | APIUserData, options: ImageURLOptions = {}) {
	if (user.avatar === null) return `${ROOT}/embed/avatars/${Number(user.discriminator) % 5}.png`;
	const format = typeof options.format === 'undefined' ? user.avatar.startsWith('a_') ? 'gif' : 'png' : options.format;
	const size = typeof options.size === 'undefined' ? '' : `?size=${options.size}`;
	return `${ROOT}/avatars/${id}/${user.avatar}.${format}${size}`;
}
```

kyranet:
`getDisplayAvatar` was meant to make getting avatars easier for UserTag and APIUserData, but any instance of User should use its respective method.

Quantumlyy (dismissed):

favna (approved):

PR #971, opened 2020-05-06
build(update): support full rebuild option

favna:
Also improved the cutting of  text that it outputs in order to hopefully finally fully get rid of exceeding 2K character limit

### DIFF
--- a/src/commands/System/Admin/update.ts
+++ b/src/commands/System/Admin/update.ts
@@ -110,7 +110,7 @@ export default class extends SkyraCommand {
 	}
 
 	private cutText(input: string, ...extraInput: string[]) {
-		const cutLength = extraInput.length ? 1950 / extraInput.length : 1950;
+		const cutLength = extraInput.length ? 1950 / (extraInput.length + 1) : 1950;
 		return cutText(input, cutLength);
 	}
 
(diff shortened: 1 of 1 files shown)


src/commands/System/Admin/update.ts
```diff
@@ -98,4 +109,9 @@ export default class extends SkyraCommand {
 		}
 	}
 
+	private cutText(input: string, ...extraInput: string[]) {
+		const cutLength = extraInput.length ? 1950 / extraInput.length : 1950;
+		return cutText(input, cutLength);
+	}
```
kyranet: You might want to check what this does, closely.

kyranet (changes requested):

src/commands/System/Admin/update.ts
```diff
@@ -98,4 +109,9 @@ export default class extends SkyraCommand {
 		}
 	}
 
+	private cutText(input: string, ...extraInput: string[]) {
+		const cutLength = extraInput.length ? 1950 / (extraInput.length + 1) : 1950;
+		return cutText(input, cutLength);
+	}
```
kyranet: What I meant is that `extraInput.length` is always 1, because you call this method always with two arguments. Also, I'm not sure what this method does, some comments would help me understand it.

kyranet (approved):

PR #976, opened 2020-05-07
feat(AutoModeration): per-monitor messages

kyranet:
This will make moderation logs more descriptive.

### DIFF
--- a/src/lib/util/Adder.ts
+++ b/src/lib/util/Adder.ts
@@ -1,3 +1,13 @@
+export class AdderError extends Error {
+
+	public amount: number;
+	public constructor(message: string, amount: number) {
+		super(message);
+		this.amount = amount;
+	}
+
+}
+
 export class Adder<T> extends Array<{ id: T; end: number }> {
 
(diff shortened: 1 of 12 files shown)


src/lib/util/Adder.ts
```diff
@@ -1,3 +1,13 @@
+export class AdderError extends Error {
+
+	public amount: number;
+	public constructor(message: string, amount: number) {
```
favna:
```suggestion
	public constructor(message: string, public amount: number) {
```

I like to keep classes that extend error for extra properties nice and tidy (yes this is valid, I learned that because Angular does this by default)

favna (changes requested):
Small change

src/lib/util/Adder.ts
kyranet: I'm confident ESLint's rules forbid that (I tried a while ago). Either way I documented the property and the constructor, which wouldn't be possible without this.

favna (approved):

PR #978, opened 2020-05-07
feat(Webhooks): added top.gg and re-balanced payments

kyranet:


src/routes/webhooks/topgg.ts
```diff
@@ -0,0 +1,37 @@
...
+import { Route, RouteOptions } from 'klasa-dashboard-hooks';
+import { TOKENS } from '@root/config';
+import { UserSettings } from '@lib/types/settings/UserSettings';
+import { ApplyOptions } from '@skyra/decorators';
+
+@ApplyOptions<RouteOptions>({ name: 'webhooks/topgg', route: 'webhooks/topgg' })
+export default class extends Route {
+
+	public async post(request: ApiRequest, response: ApiResponse) {
+		if (request.headers.authorization !== TOKENS.WEBHOOK_B4D) return response.forbidden();
```
favna:
```suggestion
		if (request.headers.authorization !== TOKENS.WEBHOOK_TOPGG) return response.forbidden();
```

Have you learned nothing from Lastpass stuff (╯°□°）╯︵ ┻━┻ unique passwords per server mate.
(config on vps has been updated)

favna (changes requested):
As per review also please update `config.example.ts`

src/routes/webhooks/topgg.ts
```diff
@@ -0,0 +1,37 @@
...
+		if (!request.body) return response.badRequest();
+
+		const body = request.body as Body;
+		try {
+			const user = await this.client.users.fetch(body.user);
+			const settings = await user.settings.sync();
+			const payment = body.isWeekend ? 600 : 400;
```
favna:
```suggestion
			const payment = body.isWeekend ? 800 : 400;
```

In the weekend user's votes count as 2 so we should also give them twice the normal amount of shinies IMO

src/routes/webhooks/topgg.ts
```diff
@@ -0,0 +1,37 @@
...
+export default class extends Route {
+
+	public async post(request: ApiRequest, response: ApiResponse) {
+		if (request.headers.authorization !== TOKENS.WEBHOOK_B4D) return response.forbidden();
+		if (!request.body) return response.badRequest();
+
+		const body = request.body as Body;
+		try {
+			const user = await this.client.users.fetch(body.user);
+			const settings = await user.settings.sync();
```
favna, line 18: While unlikely, people can vote on bots that Skyra doesn't share a server with and so will have no data for. Will `user.settings.sync()` work if there was no entry for that user in the settings? I assume fetching will succeed since it fetches from the API but I'm not sure about this.

kyranet: It does succeed, if there's nothing in the DB, it does nothing but mark the entry as "doesn't exist in DB, execute CREATE instead of UPDATE in next save".

favna (approved):

PR #991, opened 2020-05-14
fix(twitch): fixes round 2

favna:
- feat(twitch): implement twitch subscription updating
- fix(twitch): fixed online notifications
- refactor(twitch): rewrite all remaining calls to use new Helix API


kyranet (dismissed):

kyranet (dismissed):

src/lib/queries/json.ts
```diff
@@ -367,18 +372,23 @@ export class JsonCommonQuery implements CommonQuery {
 		};
 	}
 
-	public async upsertTwitchStreamSubscription(streamerID: string, guildID: string, expireSeconds = 864000) {
+	public async upsertTwitchStreamSubscription(streamerID: string, guildID?: string) {
 		const value = await this.provider.get(Databases.TwitchStreamSubscriptions, streamerID) as RawTwitchStreamSubscriptionSettings;
 		if (value) {
-			const guild_ids = value.guild_ids.concat(guildID);
-			await this.provider.update(Databases.TwitchStreamSubscriptions, streamerID, { ...value, guild_ids });
+			// When updating Twitch subscriptions the GuildID is passed as `null`
```
favna:
```suggestion
			// When updating Twitch subscriptions the GuildID is passed as `undefined`
```

src/lib/util/Notifications/Twitch.ts
```diff
@@ -27,7 +21,7 @@ export class Twitch {
 
 	public readonly ratelimitsStreams = new RateLimitManager(1, Twitch.RATELIMIT_COOLDOWN);
 	public readonly BASE_URL_HELIX = 'https://api.twitch.tv/helix/';
-	public readonly BASE_URL_KRAKEN = 'https://api.twitch.tv/kraken/';
+	public readonly brandingColour = 0x6441A4;
```
favna: rename to `BRANDING_COLOUR`

kyranet (approved):

src/events/klasaReady.ts
```diff
@@ -39,4 +40,11 @@ export default class extends Event {
 		}
 	}
 
+	private async initTwitchRefreshSubscriptionsTask() {
+		const { tasks } = this.client.schedule;
+		if (!tasks.some(task => task.taskName === 'twitchRefreshSubscriptions')) {
+			await this.client.schedule.create('twitchRefreshSubscriptions', '@daily');
```
Quantumlyy, line 46: Might want to make this more common since it doesn't really cost anything to run a DB check for the times. At the same time you will eliminate blind spots if a user expires half a day before it runs and then streams.

Quantumlyy (changes requested):

src/events/klasaReady.ts
kyranet: I think it's fine to run daily, but re-new those that are expiring in less than 24h.

favna: Any suggestions? It shouldn't be too often. DB entries are valid for 8 days from creation and Twitch subs for 9 (can be made 10).

Quantumlyy (approved):

PR #995, opened 2020-05-16
feat: pre-command logs

favna:
This is basically an identical implementation to the command logging finalizer from klasa core, but now runs as a monitor so it can run *before* commands. Lets try and hunt down that infinite loop...

### DIFF
--- a/src/monitors/preCommandHandler.ts
+++ b/src/monitors/preCommandHandler.ts
@@ -1,0 +1,43 @@
+import { ApplyOptions } from '@skyra/decorators';
+import { Colors, KlasaMessage, Monitor, MonitorOptions } from 'klasa';
+
+@ApplyOptions<MonitorOptions>({
+	ignoreBots: false,
+	ignoreEdits: false,
+	ignoreOthers: false,
+	ignoreSelf: false,
+	ignoreWebhooks: false
+})
+export default class extends Monitor {
+
+	private kDebugColours = new Colors({ background: 'black', text: 'red' });
+	private kUserColours = new Colors({ background: 'yellow', text: 'black' });
+	private kShardColours = new Colors({ background: 'cyan', text: 'black' });
+	private kChannelColours: Record<string, Colors> = {
+		text: new Colors({ background: 'green', text: 'black' }),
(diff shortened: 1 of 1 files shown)


src/monitors/preCommandHandler.ts
```diff
@@ -0,0 +1,44 @@
...
+	ignoreOthers: false,
+	ignoreSelf: false,
+	ignoreWebhooks: false
+})
+export default class extends Monitor {
+
+	private kDebugColours = new Colors({ background: 'black', text: 'red' });
+	private kUserColours = new Colors({ background: 'yellow', text: 'black' });
+	private kShardColours = new Colors({ background: 'cyan', text: 'black' });
+	private kChannelColours: Record<string, Colors> = {
```
kyranet:
```suggestion
	private readonly kDebugColours = new Colors({ background: 'black', text: 'red' });
	private readonly kUserColours = new Colors({ background: 'yellow', text: 'black' });
	private readonly kShardColours = new Colors({ background: 'cyan', text: 'black' });
	private readonly kChannelColours: Record<string, Colors> = {
```

src/monitors/preCommandHandler.ts
```diff
@@ -0,0 +1,44 @@
...
+		if (!message.command) return undefined;
+
+		const { type } = message.channel;
+		const shard = message.guild ? message.guild.shardID : 0;
+		this.client.emit('debug', [
+			this.kDebugColours.format('[Pre-Command-Run Debug Log]'),
+			this.kShardColours.format(`[${shard}]`),
+			`${message.command.name}(${message.args ? message.args.join(', ') : ''})`,
+			this.kUserColours.format(`${message.author.username}[${message.author.id}]`),
+			this.kChannelColours[type].format(type === 'text' ? this.text(message) : this.dm())
```
kyranet: This won't be accurate for the extreme case somebody runs a command in a news channel, it should check for 'dm'.
```suggestion
			this.kChannelColours[type].format(type === 'dm' ? this.dm() : this.text(message))
```

kyranet (changes requested):

kyranet (approved):

PR #1002, opened 2020-05-22
fix(Twitch): escape markdown

kyranet:


favna (changes requested):
tbh if they're in a folder called `twitch` they might as well be called `streamOnline` and `streamOffline` as files.

favna (approved):

PR #1010, opened 2020-05-30
fix(Music): escape markdown **and** mentions

kyranet:


favna (approved):

MrJacz:
this is incorrect, you should be using allowed_mentions

kyranet:
The version of d.js we use does not support allowed_mentions.

PR #1028, opened 2020-06-08
fix(C4): check must be inclusive in horizontal

kyranet:
And fixes the docs, there's not a 8th column.


favna (dismissed):
LGTM but you'll need to test it

PR #1029, opened 2020-06-09
fix(greroll): allow giveaways to be rerolled before exit, thus exit early

favna:


src/commands/Giveaway/greroll.ts
```diff
@@ -35,7 +35,9 @@ export default class extends SkyraCommand {
 
 	private async resolveMessage(message: KlasaMessage, rawTarget: KlasaMessage | undefined) {
 		const target = rawTarget
-			? this.validateMessage(rawTarget) ? rawTarget : null
+			// If rawMessage is defined then we also check for colour
+			? this.validatePossibleMessage(rawTarget) ? rawTarget : null
+			// If rawTarget was undefined then we fetch it from the API and we don't check embed colour
 			: (await message.channel.messages.fetch({ limit: 100 })).find(msg => this.validateMessage(msg)) || null;
```
kyranet:
```suggestion
			// If rawMessage is defined then we check everything sans the colour
			? this.validateMessage(rawTarget) ? rawTarget : null
			// If rawTarget was undefined then we fetch it from the API and we check embed colour
			: (await message.channel.messages.fetch({ limit: 100 })).find(msg => this.validatePossibleMessage(msg)) || null;
```

kyranet (changes requested):
You got that backwards.

kyranet (approved):

PR #1040, opened 2020-06-17
fix(ship): fixed for when user is (magically) undefined

favna:
Some odd edge case that occurrs in Ship sometimes apparently.

### DIFF
--- a/src/commands/Misc/ship.ts
+++ b/src/commands/Misc/ship.ts
@@ -1,34 +1,28 @@
-import { SkyraCommand } from '@lib/structures/SkyraCommand';
+import { SkyraCommand, SkyraCommandOptions } from '@lib/structures/SkyraCommand';
 import { CanvasColors } from '@lib/types/constants/Constants';
 import { UserSettings } from '@lib/types/settings/UserSettings';
 import { KeyedMemberTag } from '@root/arguments/membername';
+import { ApplyOptions, CreateResolvers } from '@skyra/decorators';
 import { socialFolder } from '@utils/constants';
 import { fetch, FetchResultTypes, getDisplayAvatar, loadImage } from '@utils/util';
 import { Image } from 'canvas';
 import { Canvas } from 'canvas-constructor';
 import { remove as removeConfusables } from 'confusables';
-import { CommandStore, KlasaMessage } from 'klasa';
+import { KlasaMessage } from 'klasa';
 import { join } from 'path';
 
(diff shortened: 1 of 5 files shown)


kyranet (dismissed):
This is a wrong fix. It should be handled wherever this is called and never be called with null or undefined.

src/commands/Misc/ship.ts
```diff
@@ -53,8 +53,11 @@ export default class extends SkyraCommand {
 
 	public async run(message: KlasaMessage, [firstUser, secondUser]: [KeyedMemberTag, KeyedMemberTag]) {
 		// We need the UserTags to resolve the avatar and username
-		const firstUserTag = this.client.userTags.get(firstUser.id)!;
-		const secondUserTag = this.client.userTags.get(secondUser.id)!;
+		let firstUserTag = this.client.userTags.get(firstUser.id);
+		let secondUserTag = this.client.userTags.get(secondUser.id);
+
+		if (!firstUserTag) firstUserTag = { avatar: null, discriminator: '0001', bot: false, username: '' };
+		if (!secondUserTag) secondUserTag = { avatar: null, discriminator: '0001', bot: false, username: '' };
```
kyranet:
```suggestion
		const firstUserTag = this.client.userTags.get(firstUser.id) ?? { avatar: null, discriminator: '0001', bot: false, username: '' };
		const secondUserTag = this.client.userTags.get(secondUser.id) ?? { avatar: null, discriminator: '0001', bot: false, username: '' };
```

kyranet: Or even better:

```ts
private readonly kEmptyUserTag = { avatar: null, discriminator: '0001', bot: false, username: '' };
```

And then:

```suggestion
		const firstUserTag = this.client.userTags.get(firstUser.id) ?? this.kEmptyUserTag;
		const secondUserTag = this.client.userTags.get(secondUser.id) ?? this.kEmptyUserTag;
```

kyranet (dismissed):

favna:
So about the ship problem @kyranet, the empty user tag thing wouldn't really work because the usertag is also being read for the  text and such. I have 2 ideas rn
1. (currently coded): pick a random usertag based on a randomKey of the memberTags, to ensure that the random usertag is in the current guild at least.
2. default to skyra lol

kyranet:
3. use fetch instead of get

kyranet (approved):

PR #1041, opened 2020-06-17
fix(gsearch): validate that image has a valid image url before adding to embed

favna:


src/commands/Google/gsearch.ts
```diff
@@ -45,11 +45,19 @@ export default class extends RichDisplayCommand {
...
+			display.addPage((embed: MessageEmbed) => {
+				embed
+					.setTitle(item.title)
+					.setURL(item.link)
+					.setDescription(item.snippet);
+
+				const imageUrl = item.pagemap?.cse_image?.find(image => parseURL(image.src))?.src ?? '';
+				if (IMAGE_EXTENSION.test(imageUrl)) {
+					embed.setImage(imageUrl);
+				}
```
kyranet: The test should be done inside the find.

kyranet (dismissed):

kyranet (approved):

PR #1042, opened 2020-06-17
fix(marry): fix 2k char limit in display

favna:
idk what else it could be other than the question. Everything else is well within 2K characters that Skyra sends in this question. So I guess cutting that one then...

See: https://discordapp.com/channels/541738403230777351/648662983324729375/712839236792877126


kyranet (commented):
A good fix would be to turn it into different pages.
Once I do `@skyra/anti-user-settings` and we migrate Skyra to TypeORM, we'll limit marriages to 20 users.

favna:
image

src/commands/Social/marry.ts
```diff
@@ -66,6 +85,7 @@ export default class extends SkyraCommand {
 
 		// settings is already sync by the monitors.
 		const spouses = author.settings.get(UserSettings.Marry);
+		this.client.console.log('spouses: ', spouses);
```
kyranet: 🧹
```suggestion
```

kyranet (changes requested):

src/commands/Social/marry.ts
```diff
@@ -24,36 +29,51 @@ async function askYesNo(channel: TextChannel | DMChannel, user: KlasaUser, quest
...
+
 		const users = message.author.settings.get(UserSettings.Marry);
 		if (users.length === 0) return message.sendLocale('COMMAND_MARRY_NOTTAKEN');
 
-		const usernames = await Promise.all(users.map(async user => `${await this.client.userTags.fetchUsername(user)} (${user})`));
-		return message.sendLocale('COMMAND_MARRY_WITH', [usernames]);
+		const usernames = chunk(await Promise.all(users.map(async user => {
+			const [userId] = await this.client.userTags.fetchEntry(user);
+			return `<@${userId}>`;
+		})), 20);
```
kyranet: Are you not going to use the username? It was added because sometimes mentions don't render well.
If not, change it to this:

```suggestion
		const usernames = chunk(users.map(user => `<@${user}>`), 20);
```

kyranet (dismissed):

src/commands/Social/marry.ts
favna: in what cases would mentions not render well? Is this when they are married to someone they no longer share a server with?

kyranet: Yup. Or even when the married users are uncached on the user's side.

favna: Alright I'll revert this later today then. Currently updating pokemon API with DLC data.

kyranet (approved):

PR #1057, opened 2020-06-21
Fix spellling mistake inside the README.md file.

Tylertron1998:


favna (approved):

favna:
Only because apparently `doughnut` is more common in writing, even in American English. source: https://writingexplained.org/doughnut-or-donut-difference

PR #1082, opened 2020-07-14
feat(TORM): Command usage analytics migrations

Quantumlyy:


src/lib/orm/migrations/1594757329224-V13_MigrateAnalytics.ts
```diff
@@ -0,0 +1,105 @@
...
+					INSERT
+					INTO command_counter ("id", "uses")
+					VALUES ($1, $2)
+					ON CONFLICT (id)
+					DO
+						UPDATE
+						SET uses = command_counter.uses + $2;
+				`, [command._field, command._value]);
+			}
+		});
```
kyranet: No need for the `.then()` in an `async/await`, also you probably want to wrap it into a `try/catch/finally` so you commit on success, rollback on failure (and rethrow), and dispose when done.

src/lib/orm/migrations/1594757329224-V13_MigrateAnalytics.ts
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	public async down(queryRunner: QueryRunner): Promise<void> {
+		const influx = new InfluxDB(INFLUX_OPTIONS);
+		const reader = influx.getQueryApi(INFLUX_ORG);
+
+		const commands = await reader.collectRows<InfluxSummedCommandEntry>(INFLUX_ALL_COMMANDS_SCRIPT);
+
+		await queryRunner.startTransaction().then(async () => {
+			await queryRunner.query('CREATE TABLE command_counter');
```
kyranet: You should use `queryRunner.createTable` and follow what we did in other migrations, this syntax is invalid as it doesn't include the columns.

kyranet (changes requested):

kyranet (dismissed):

src/lib/orm/migrations/1594757329224-V13_MigrateAnalytics.ts
```diff
@@ -0,0 +1,120 @@
+import { InfluxDB, Point, WritePrecision } from '@influxdata/influxdb-client';
```
favna, line 1: Have you locally tested the migration? If not please do so, you can get into a pre-migration state with the `typeorm:schema:resync` script seeing as how this PR does not yet remove the entity or any calls to it.

favna:
Once #1090 is merged please also add a line to the SQL script that adds the new migration to the table (SQL is pretty self explanatory there). The SQL script is to fake having ran migrations.

src/lib/orm/migrations/1594757329224-V13_MigrateAnalytics.ts
Quantumlyy: Well, it seems to have worked now.

favna:
well there's 1 test you will have to do still
1. checkout some commit or branch that is pre-typeorm (for a branch I guess you could grab misc/moderations-reformat)
2. delete all your database tables and the rpg enum thingy
3. build & run the bot so it creates the tables with Klasa SG
4. Do some commands to fill command counter
5. Stop the bot
6. Run the migrations

That will basically simulate the state from which migrations will be ran on prod.

Quantumlyy:
> well there's 1 test you will have to do still
> 
> 1. checkout some commit or branch that is pre-typeorm (for a branch I guess you could grab misc/moderations-reformat)
> 2. delete all your database tables and the rpg enum thingy
> 3. build & run the bot so it creates the tables with Klasa SG
> 4. Do some commands to fill command counter
> 5. Stop the bot
> 6. Run the migrations
> 
> That will basically simulate the state from which migrations will be ran on prod.

This test has been performed successfully, both for run and revert.

favna (approved):

kyranet (approved):

PR #1090, opened 2020-07-15
refactor/scripts stuff

favna:
- chore: remove Database folder
- chore: add setmigrations sql script
- docs: add doc and support for .mjs scripts in scripts folder
- docs: add @Rexogamer as a contributor


scripts/tsconfig.json
```diff
@@ -0,0 +1,13 @@
...
+		"target": "ESNext",
+		"module": "ESNext",
+		"rootDir": ".",
+		"outDir": ".",
+		"sourceMap": false,
+		"incremental": false
+	},
+	"include": [".", "*.mjs"],
+	"exclude": []
+}
```
kyranet:
```suggestion
}

```

scripts/gulpfile.ts
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	@Task('subTask_removeJsOutput')
+	public async subTask_removeJsOutput() {
+		return (await fsp.readdir(this.#scriptsDir))
+			.filter(file => this.#jsFilesPattern.test(file))
+			.map(file => remove(resolve(this.#scriptsDir, file)));
+	}
+
+
```
kyranet:
```suggestion

```

scripts/gulpfile.ts
```diff
@@ -0,0 +1,90 @@
...
+import rename from 'gulp-rename';
+import { Gulpclass, SequenceTask, Task } from 'gulpclass/Decorators';
+import fetch from 'node-fetch';
+import { join, resolve } from 'path';
+import { toUnicode } from 'punycode';
+import { createConnection } from 'typeorm';
+
+// Get the TypeORM configuration
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const config = require('./ormconfig');
```
kyranet: You probably want to specify that it doesn't refer to an `ormconfig` at the `scripts` folder, but that it's `eval`uated in the root folder (`gulpfile.js`) after transpiling it from TypeScript.

Though I'm not fond of this practice and feels too hacky.

kyranet (changes requested):

scripts/gulpfile.ts
favna: It's either that or using ts-node (which shits on reading tsconfig and is bad anyway) or writing the gulp tasks in pure JS.

Or I suppose adding individual JS files in scripts but then we lose typechecking.

Initially I had it set up with just .mjs files but the lack of typechecking bothers me.

kyranet: > It's either that or using ts-node (which shits on reading tsconfig and is bad anyway) or writing the gulp tasks in pure JS.

What I meant is to place a line specifying this file is run in another context, aka it runs in the root folder as the entire module's source code is run via `eval()`. This isn't obvious and will cause confusion in the future.

> Or I suppose adding individual JS files in scripts

Honestly I prefer this over having to install multiple dependencies, aligns scripts to how we do the tests, bringing consistency to the project (and avoiding `gulp`, a module I'm not exactly happy with).

> Initially I had it set up with just .mjs files but the lack of typechecking bothers me.

`tsconfig.json` can be configured to typecheck JavaScript's types, that's how [Skyra 3.4.0 was typechecked](https://github.com/skyra-project/skyra/blob/3.4.0/tsconfig.json#L9) despite of being entirely in JavaScript.

kyranet (approved):

PR #1091, opened 2020-07-15
fix(rpg): change RPGRank#guild to a ManyToOne relationship

kyranet:
Fixes a bug introduced by the automatic migration tool for SQL to TypeORM.


PyroTechniac (dismissed):

src/lib/orm/entities/RpgGuildRankEntity.ts
```diff
@@ -12,7 +12,7 @@ export class RpgGuildRankEntity extends BaseEntity {
 	@Column('varchar', { length: 50 })
 	public name!: string;
 
-	@OneToMany(() => RpgGuildEntity, guild => guild.ranks, { onDelete: 'CASCADE' })
+	@ManyToOne(() => RpgGuildEntity, guild => guild.ranks, { onDelete: 'CASCADE' })
```
PyroTechniac, line 15: I don't think the ManyToOne side needs to show the inverse

kyranet: All other references of `@ManyToOne` in the project have an inverse, I'd keep this for consistency.

But if you can prove the behaviour is the same without them, feel free to submit a PR later 👍

favna (approved):

PR #1118, opened 2020-07-19
refactor(sticky-roles): centralize all logic and fix everything

kyranet:


src/lib/structures/managers/StickyRoleManager.ts
```diff
@@ -0,0 +1,117 @@
...
+	public async fetch(userID: string, extraContext?: unknown): Promise<readonly string[]> {
+		const arrayIndex = this.getKey(userID);
+		if (arrayIndex === -1) return [];
+
+		const entry = this.entries[arrayIndex];
+		const roles = [...this.cleanRoles(entry.roles)];
+		if (entry.roles.length === roles.length) return entry.roles;
+
+		const clone: StickyRole = { user: userID, roles };
+		await this.#guild.settings.update(GuildSettings.StickyRoles, clone, { arrayIndex, extraContext });
```
favna: For consistency please wrap this over multiple lines like in other places where update is used

src/lib/structures/managers/StickyRoleManager.ts
```diff
@@ -0,0 +1,117 @@
...
+		return roles;
+	}
+
+	public async add(userID: string, roleID: string, extraContext?: unknown): Promise<readonly string[]> {
+		const arrayIndex = this.getKey(userID);
+		if (arrayIndex === -1) {
+			const entry: StickyRole = {
```
favna: To avoid confusion of other variables also called entry in this file with wholly different context I suggest renaming this variable

src/lib/structures/managers/StickyRoleManager.ts
```diff
@@ -0,0 +1,117 @@
...
+				user: userID,
+				roles: [roleID]
+			};
+			await this.#guild.settings.update(GuildSettings.StickyRoles, entry, { arrayAction: 'add', extraContext });
```
favna: `arrayAction: ArrayActions.Add` use the enum from klasa

src/lib/structures/managers/StickyRoleManager.ts
```diff
@@ -0,0 +1,117 @@
...
+
+	private get entries() {
+		return this.#guild.settings.get(GuildSettings.StickyRoles);
+	}
+
+	public get(userID: string): readonly string[] {
+		return this.entries.find(entry => entry.user === userID)?.roles ?? [];
+	}
+
+	public async fetch(userID: string, extraContext?: unknown): Promise<readonly string[]> {
```
favna: extraContent should be typed like this because we're always passing in the author (and if we're not then we should)

```ts
type ExtraContent<T> = T & { author: string }
```

favna (changes requested):

src/serializers/stickyrole.ts
```diff
@@ -10,14 +10,7 @@ export default class extends Serializer {
...
 			&& Array.isArray(data.roles)
-			&& data.roles.every(role => typeof role === 'string')) {
-
-			// TODO: Resolve this to a better solution (i.e. using `guild.roles.sticky.add`)
-			// see https://discordapp.com/channels/541738403230777351/541740581832097792/734011404750684220
-			Reflect.set(data, 'roles', [...new Set(data.roles)].filter((role => guild!.roles.has(role))));
-			return data;
-
-		}
+			&& data.roles.every(role => typeof role === 'string' && guild!.roles.has(role))) return data;
```
favna, line 13: What will happen if the user has a sticky role that has been removed now that you revert this...

kyranet: It's covered by the `StickyRoleManager`, if you read the comments, it's theoretically impossible for that to ever return false.

This check is there so people don't push a StickyRole entry with +1000 invalid roles via the dashboard's API.

PR #1120, opened 2020-07-19
fix(starboard): resolved bug with the system not updating

kyranet:


src/lib/orm/entities/StarboardEntity.ts
```diff
@@ -180,8 +178,13 @@ export class StarboardEntity extends BaseEntity {
 		const previousUpdate = this.#manager.syncMessageMap.get(this);
 		if (previousUpdate) await previousUpdate;
 
-		if ('enabled' in options) this.enabled = options.enabled!;
-		if ('stars' in options && this.enabled) await this.updateStarMessage();
+		if ('enabled' in options) {
```
favna, line 181: Didn't we agree we'd prefer reflect syntax? I thought we did 🙄. I mean it's not that big a deal but...

kyranet: We're switching to the type safer `isSet` from klasa-core when we do the switch.
Anyways, all the ` in ` refactors can be done later in another PR, I kept the original here as refactoring it is out of the PR's scope.

favna (approved):

PR #1123, opened 2020-07-19
feat(Analytics): Resources

Quantumlyy:


src/events/klasaReady.ts
```diff
@@ -58,11 +58,20 @@ export default class extends Event {
 		}
 	}
 
-	private initAnalytics() {
+	private async initSyncResourceAnalyticsTask() {
+		const { queue } = this.client.schedules;
+		if (!queue.some(task => task.taskID === Schedules.SyncResourceAnalytics)) {
+			await this.client.schedules.add(Schedules.TwitchRefreshSubscriptions, '*/1 * * * *');
```
kyranet:
```suggestion
		if (!queue.some(task => task.taskID === Schedules.SyncResourceAnalytics)) {
			await this.client.schedules.add(Schedules.SyncResourceAnalytics, '*/1 * * * *');
```

src/events/analytics/resourceAnalyticsSync.ts
```diff
@@ -0,0 +1,41 @@
...
+
+		return point;
+	}
+
+	private syncMem() {
+		const usage = process.memoryUsage();
+		return new Point(AnalyticsSchema.Points.Memory)
+			.tag(AnalyticsSchema.Tags.Action, AnalyticsSchema.Actions.Sync)
+			.floatField('total', Math.round(100 * (usage.heapTotal / 1048576)) / 100)
+			.floatField('used', Math.round(100 * (usage.heapUsed / 1048576)) / 100);
```
kyranet: Couldn't those just return raw values? We can map them in InfluxDB to do the division and the decimal trimming.

kyranet (changes requested):

src/events/analytics/resourceAnalyticsSync.ts
Quantumlyy: Seems like I missed them whilst converting CPU to raw values.

kyranet (approved):

favna (approved):

PR #1125, opened 2020-07-20
fix: handle music join exceptions

favna:
For some bizarre reason people were using "disconnect" for the bot which puts it in a half state of playing and not playing so instead we'll just call a full leave when an exception occurs.

### DIFF
--- a/src/lib/structures/music/MusicHandler.ts
+++ b/src/lib/structures/music/MusicHandler.ts
@@ -1,5 +1,6 @@
 import { SkyraClient } from '@lib/SkyraClient';
 import { Events } from '@lib/types/Enums';
+import { LavalinkPlayerEvents } from '@lib/types/Events';
 import { GuildSettings } from '@lib/types/settings/GuildSettings';
 import { flattenMusicHandler } from '@utils/Models/ApiTransform';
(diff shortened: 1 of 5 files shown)


src/lib/structures/music/MusicHandler.ts
```diff
@@ -266,6 +267,14 @@ export class MusicHandler {
 		if (volume) this.volume = this.guild.settings.get(GuildSettings.Music.DefaultVolume);
 	}
 
+	public async handleException(context: MusicHandlerRequestContext | null = null) {
+		// Destroy the lavalink instance
+		await this.leave(context);
+
+		// Empty the queue
+		this.queue = [];
+	}
```
kyranet: Is it necessary to do this? Specially clearing up the queue?
LL could return an exception because one track is not allowed in a country and Skyra would immediately leave and empty the entire queue.

kyranet (dismissed):

src/lib/structures/music/MusicHandler.ts
favna: I'll handle it in another place, had a better idea.

kyranet (approved):
Pretty solid code

PR #1132, opened 2020-07-22
fix(apiresponse): set CORS headers on API responses

favna:


kyranet (commented):
This will cause issues as you'd be sending the same 3 headers twice:

https://github.com/skyra-project/skyra/blob/master/src/middlewares/headers.ts

favna:
Well then that middleware doesn't work because it was saying that I didn't set the header at all before...

favna:
dafuq now it does work

PR #1135, opened 2020-07-24
feat(tests/coverage): use v8 over babel

Quantumlyy:


jest.config.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
+	coverageProvider: 'v8',
```
favna, line 2:
```suggestion
	coverageProvider: 'v8',
	coverageReporters: ['lcov', 'text-summary'],
```

Not saying this will fix the failing CI but just maybe it will. Lcov is needed for coveralls, text-summary is for STDOUT.

favna (changes requested):

favna:
Alas it looks like V8 coverage reports break our unit tests in this project so I'm closing this. Maybe we can come back to this in a later version of Jest...

PR #1136, opened 2020-07-24
feat: define command

adiologydev:
This PR adds definition command back to Skyra which is now utilizing Owlbot's API.

- Added `define.ts`
- Update `en-US` and `es-ES`
- Update config and types


src/commands/Tools/Dictionary/define.ts
```diff
@@ -0,0 +1,82 @@
...
+import { DbSet } from '@lib/structures/DbSet';
+import { RichDisplayCommand, RichDisplayCommandOptions } from '@lib/structures/RichDisplayCommand';
+import { UserRichDisplay } from '@lib/structures/UserRichDisplay';
+import { cutText, fetch, FetchResultTypes } from '@utils/util';
+import { toTitleCase } from '@klasa/utils';
+import { MessageEmbed } from 'discord.js';
+import { ApplyOptions } from '@skyra/decorators';
+import { BrandingColors } from '@utils/constants';
+
+const options = { headers: { Accept: 'application/json', Authorization: `Token ${TOKENS.OWLBOT}` } };
```
kyranet: `'application/json'` -> `Mime.Types.ApplicationJson`

favna: Just inline the options below, We don't really set it outside in other commands either. Also as Kyra said in Discord, use the Mime.Types enum for the accept.

src/commands/Tools/Dictionary/define.ts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+export interface OwlbotResultOk {
+	definitions: Array<{
+		type: string;
+		definition: string;
+		example: string|null;
+		image_url: string|null;
+		emoji: string|null;
+	}>;
```
kyranet: The contents of definitions should be its own interface, and then change this to

```ts
definitions: readonly OwlbotResultDefinition[];
```

kyranet (changes requested):

src/commands/Tools/Dictionary/define.ts
```diff
@@ -0,0 +1,82 @@
...
+				options,
+				FetchResultTypes.JSON
+			);
+
+			const display = await this.buildDisplay(result, message);
+			await display.start(response, message.author.id);
+			return response;
+		} catch {
+			throw message.language.get('COMMAND_DEFINE_NOTFOUND');
+		}
```
favna: Move the fetch to a fetchApi function because rn api errors affect message sending. 

```ts
const result = await this.fetchApi(...);
```

src/commands/Tools/Dictionary/define.ts
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	private async fetchApi(message: KlasaMessage, word: string) {
+		try {
+			const res = await fetch<OwlbotResultOk>(
+				`https://owlbot.info/api/v4/dictionary/${encodeURIComponent(word.toLowerCase())}`,
+				{ headers: { Accept: Mime.Types.ApplicationJson, Authorization: `Token ${TOKENS.OWLBOT}` } },
+				FetchResultTypes.JSON
+			);
+			return res;
```
favna:
```suggestion
			return await fetch<OwlbotResultOk>(
				`https://owlbot.info/api/v4/dictionary/${encodeURIComponent(word.toLowerCase())}`,
				{ headers: { Accept: Mime.Types.ApplicationJson, Authorization: `Token ${TOKENS.OWLBOT}` } },
				FetchResultTypes.JSON
			);
```

src/commands/Tools/Dictionary/define.ts
```diff
@@ -0,0 +1,87 @@
...
+	usage: '<input:string>'
+})
+export default class extends RichDisplayCommand {
+
+	public async run(message: KlasaMessage, [input]: [string]) {
+		const response = await message.sendEmbed(new MessageEmbed()
+			.setDescription(message.language.tget('SYSTEM_LOADING'))
+			.setColor(BrandingColors.Secondary));
+
+		const result = await this.fetchApi(message, input);
```
favna, line 27: What does the API return when there is no definition? 400/500 errors won't be caught by the catch. If so then handle it here. See for example https://github.com/skyra-project/skyra/blob/9a6b1d0fe27c82f9b7ed461bed724d750e56cd12/src/commands/Anime/anime.ts#L30-L35

src/commands/Tools/Dictionary/define.ts
```diff
@@ -0,0 +1,87 @@
...
+		const display = await this.buildDisplay(result, message);
+		await display.start(response, message.author.id);
+		return response;
+
```
favna:
```suggestion
		const display = await this.buildDisplay(result, message);
		
		await display.start(response, message.author.id);
		return response;
```

favna (changes requested):

src/commands/Tools/Dictionary/define.ts
adiologydev: It actually does catch 400/500 errors because I'm awaiting the response within the method as it's assigned to the variable.

Thus it'll catch the error, also I've tested it.

src/commands/Tools/Dictionary/define.ts
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	private async buildDisplay(results: OwlbotResultOk, message: KlasaMessage) {
+		const display = new UserRichDisplay(new MessageEmbed()
+			.setTitle(toTitleCase(results.word))
+			.setColor(await DbSet.fetchColor(message))
+			.addField('Pronounciation', results.pronunciation, true))
```
kyranet: Move `'Pronounciation'` to i18n

kyranet (changes requested):

src/commands/Tools/Dictionary/define.ts
```diff
@@ -0,0 +1,90 @@
...
+		if (definition.length < 2048) return definition;
+		return cutText(definition, 2048);
+	}
+
+}
+
+export interface OwlbotResultOk {
+	definitions: readonly OwlbotDefinition[];
+	word: string;
+	pronunciation: string|null;
```
kyranet:
```suggestion
	pronunciation: string | null;
```

src/commands/Tools/Dictionary/define.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+export interface OwlbotDefinition {
+	type: string;
+	definition: string;
+	example: string|null;
+	image_url: string|null;
+	emoji: string|null;
```
kyranet:
```suggestion
	example: string | null;
	image_url: string | null;
	emoji: string | null;
```

kyranet (changes requested):
LGTM when resolved

src/commands/Tools/Dictionary/define.ts
```diff
@@ -0,0 +1,90 @@
...
+		const response = await message.sendEmbed(new MessageEmbed()
+			.setDescription(message.language.tget('SYSTEM_LOADING'))
+			.setColor(BrandingColors.Secondary));
+
+		const result = await this.fetchApi(message, input);
+		const display = await this.buildDisplay(result, message);
+
+		await display.start(response, message.author.id);
+		return response;
+
```
favna:
```suggestion
```

🧹

kyranet (dismissed):

src/commands/Tools/Dictionary/define.ts
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	private async buildDisplay(results: OwlbotResultOk, message: KlasaMessage) {
+		const base = new MessageEmbed()
+			.setTitle(toTitleCase(results.word))
+			.setColor(await DbSet.fetchColor(message));
+
+		if (results.pronunciation) base.addField(message.language.get('COMMAND_DEFINE_PRONOUNCIATION'), results.pronunciation, true);
+
+		const display = new UserRichDisplay(base)
```
favna:
```suggestion
		const template = new MessageEmbed()
			.setTitle(toTitleCase(results.word))
			.setColor(await DbSet.fetchColor(message));

		if (results.pronunciation) template.addField(message.language.get('COMMAND_DEFINE_PRONOUNCIATION'), results.pronunciation, true);

		const display = new UserRichDisplay(template)
```

src/commands/Tools/Dictionary/define.ts
```diff
@@ -0,0 +1,90 @@
...
+			.setFooterSuffix(' - Powered by Owlbot');
+
+		for (const result of results.definitions) {
+			const definition = this.content(result.definition);
+			display.addPage((embed: MessageEmbed) => {
+				embed
+					.addField('Type', toTitleCase(result.type), true)
+					.setDescription(`${definition[0].toUpperCase()}${definition.substr(1)}`);
+				if (result.image_url) embed.setThumbnail(result.image_url);
```
favna:
```suggestion
				if (result.image_url) embed.setThumbnail(result.image_url);
```

if `gimage` and `gsearch` are anything to go by then relying on API's to give proper proper image URLs is a lost fight. Can you check `gimage`'s way of validating the image is valid before attaching it and implement that here too?

favna (changes requested):

src/languages/en-US.ts
```diff
@@ -3901,6 +3901,16 @@ export default class extends Language {
...
+		COMMAND_DEFINE_DESCRIPTION: 'Check the definition of a word.',
+		COMMAND_DEFINE_EXTENDED: builder.display('define', {
+			extendedHelp: `What does "heel" mean?`,
+			explainedUsage: [
+				['Word', 'The word or phrase you want to get the definition from.']
+			],
+			examples: ['heel']
+		}),
+		COMMAND_DEFINE_NOTFOUND: 'I could not find a definition for this word.',
+		COMMAND_DEFINE_PRONOUNCIATION: 'Pronounciation',
```
Rexogamer: Slight typo:
```suggestion
		COMMAND_DEFINE_PRONOUNCIATION: 'Pronunciation',
```

Rexogamer:
Also: should `def` be an alias for this command?

favna (approved):

kyranet (approved):

PR #1156, opened 2020-07-31
fix(dog): command erroring

kyranet:


src/commands/Animal/dog.ts
```diff
@@ -16,23 +16,23 @@ import { KlasaMessage } from 'klasa';
...
+			.setTimestamp());
+	}
 
-		return message.sendEmbed(embed);
+	private async fetchImage() {
+		const randomDogData = await fetch<DogResultOk>('https://dog.ceo/api/breeds/image/random', FetchResultTypes.JSON)
+			.catch(() => null);
+		return randomDogData?.status === 'success' && isImageURL(randomDogData.message)
+			? randomDogData.message
+			: 'https://i.imgur.com/cF0XUF5.jpg';
```
favna, line 35: Should make this a try-catch, to match commands like anime (where you made that change from this structure in the typorm branch)

favna (changes requested):

src/commands/Animal/dog.ts
kyranet: It's intentional to use `.catch(() => null)` because we had a `if (randomDogData && randomDogData.status === 'success'` before, in case `fetch` returned unexpected data (happened before tbh).

Making the catch like that allows me to handle both cases in the same if, thus making it much clearer.

favna: ah, fair enough.

favna (approved):

PR #1160, opened 2020-07-31
fix(wof): floor/ceil floats

kyranet:


src/lib/util/Games/WheelOfFortune.ts
```diff
@@ -108,8 +108,8 @@ export class WheelOfFortune {
 
 		// The winnings
 		this.winnings = multiplier >= 1
-			? (this.amount * multiplier * await this.fetchBoost()) - this.amount
-			: (this.amount * multiplier) - this.amount;
+			? Math.floor((this.amount * multiplier * await this.fetchBoost()) - this.amount)
+			: Math.ceil((this.amount * multiplier) - this.amount);
```
favna: Probably should use `roundNumber()` from utils here instead, it defaults to 0 decimals iirc so you can omit second parameter.

favna (changes requested):

src/lib/util/Games/WheelOfFortune.ts
kyranet: Fixed in df1a4b2c

favna (approved):

PR #1161, opened 2020-07-31
feat(markov): capture all content

### LINKED ISSUE #881: request: Markov changes
- [x] Custom emojis aren't showing
- [x] Should ignore URLs as those produce buggy output
- [x] Should get full message content, i.e embed content included

kyranet:
Fixes #881

### DIFF
--- a/src/lib/util/util.ts
+++ b/src/lib/util/util.ts
@@ -346,6 +346,24 @@ export function getContent(message: Message): string | null {
 	return null;
 }
 
+/**
+ * Gets all the contents from a message.
+ * @param message The Message instance to get all contents from
+ */
+export function getAllContent(message: Message): string {
+	const output: string[] = [];
+	if (message.content) output.push(message.content);
+	for (const embed of message.embeds) {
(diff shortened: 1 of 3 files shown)


src/lib/util/util.ts
```diff
@@ -346,6 +346,24 @@ export function getContent(message: Message): string | null {
 	return null;
 }
 
+/**
+ * Gets all the contents from a message.
+ * @param message The Message instance to get all contents from
+ */
+export function getAllContent(message: Message): string {
+	const output: string[] = [];
+	if (message.content) output.push(message.content);
```
favna, line 355: Don't we want cleanContent here? Or call cleanMentions otherwise?  Same for embed descriptions and field values concerning mentions.

kyranet: No, this will be used for moderation monitors and markov, they never echo to a message content.

favna: kk

favna (approved):

PR #1174, opened 2020-08-02
feat(volume): add easter egg with an increased volume

kyranet:
I'll be honest, volume 300 sounds horrible, but it was requested and I will give them this surprise whenever they try it.


favna (dismissed):
people are total nutters

favna (approved):

PR #1181, opened 2020-08-03
feat: trivia command

Stitch07:


package.json
```diff
@@ -67,6 +67,7 @@
 		"fs-nextra": "^0.5.1",
 		"gifencoder": "^2.0.1",
 		"he": "^1.2.0",
+		"html-entities": "^1.3.1",
```
kyranet: We already have [`he`](https://www.npmjs.com/package/he) which removes HTML entities.

src/commands/GameIntegration/clashofclans.ts
```diff
@@ -62,8 +62,9 @@ export default class extends RichDisplayCommand {
 
 	public async player(message: KlasaMessage, [player]: [string]) {
 		const playerData = await this.fetchAPI<ClashOfClansFetchCategories.PLAYERS>(message, player, ClashOfClansFetchCategories.PLAYERS);
-
-		return message.send(await this.buildPlayerEmbed(message, playerData));
+		const embed = await this.buildPlayerEmbed(message, playerData);
+		console.log(embed);
+		return message.send(embed);
```
kyranet: 🧹
```suggestion
		return message.send(await this.buildPlayerEmbed(message, playerData));
```

src/lib/util/Games/TriviaManager.ts
```diff
@@ -0,0 +1,77 @@
+/* eslint-disable @typescript-eslint/no-use-before-define */
```
kyranet, line 1: Please unit test this (so coveralls doesn't complain). You can use `nock` for this (and I think it's already installed).

src/lib/util/util.ts
```diff
@@ -711,6 +711,22 @@ export function isTextBasedChannel(channel: Channel) {
...
+	const clone = array.slice();
+	const shuffled = [] as T[];
+
+	for (const { } of array) {
+		const [value] = clone.splice(Math.floor(Math.random() * clone.length), 1);
+		shuffled.push(value);
+	}
+
+	return shuffled;
+};
```
kyranet: You should use this, it's more efficient, and once you do that, you can make HG use this utility in the next PR:

https://github.com/skyra-project/skyra/blob/475551c3b01b9b34cc655e01aa71b7a365a42e30/src/commands/Games/hungergames.ts#L167-L171

kyranet (changes requested):

src/lib/util/util.ts
kyranet: Also unit test this utility.

package.json
tech6hutch: > 
> 
> We already have [`he`](https://www.npmjs.com/package/he) which removes HTML entities.

Can we get `she` as an alias

src/commands/Games/trivia.ts
```diff
@@ -0,0 +1,88 @@
...
+})
+export default class extends SkyraCommand {
+
+	// eslint-disable-next-line @typescript-eslint/explicit-member-accessibility
+	#channels = new Set<string>();
+
+	public async run(message: KlasaMessage, [category, questionType = undefined, difficulty = undefined, duration = 30]: [number, QuestionType?, QuestionDifficulty?, number?]) {
+		if (this.#channels.has(message.channel.id)) throw message.language.tget('COMMAND_TRIVIA_ACTIVE_GAME');
+
+		this.#channels.add(message.channel.id);
```
kyranet, line 24: Please dispose this data with a try/finally. Lines 25, 26, and 32 can throw errors.

src/commands/Games/trivia.ts
```diff
@@ -0,0 +1,88 @@
...
+		await message.sendLocale('SYSTEM_LOADING');
+		const data = await TriviaManager.getQuestion(category, difficulty, questionType);
+		const possibleAnswers = questionType === QuestionType.BOOLEAN
+			? ['True', 'False']
```
kyranet: Move to i18n, as well as make them lower-case so we can match them with `.toLowerCase()`.

src/commands/Games/trivia.ts
```diff
@@ -0,0 +1,88 @@
...
+			: shuffle([data.correct_answer, ...data.incorrect_answers].map(ans => decode(ans)));
+		const correctAnswer = decode(data.correct_answer);
+
+		await message.sendEmbed(this.buildQuestionEmbed(message, data, possibleAnswers));
+		const filter = (msg: KlasaMessage) => {
+			const num = parseInt(msg.content, 10);
```
kyranet: We always use `Number`, since `parseInt('11aa');` returns `11` instead of `NaN`.

src/commands/Games/trivia.ts
```diff
@@ -0,0 +1,88 @@
...
+				questionDisplay.join('\n')
+			].join('\n'));
+	}
+
+	public async init() {
+		this.createCustomResolver('category', (arg, possible, message) => {
+			if (!arg) return CATEGORIES.general;
+			arg = arg.toLowerCase();
+			if (!Reflect.has(CATEGORIES, arg)) throw message.language.tget('COMMAND_TRIVIA_INVALID_CATEGORY');
+			return Reflect.get(CATEGORIES, arg);
```
kyranet: has + get is an anti-pattern, get once and check its value.

src/lib/util/Games/TriviaManager.ts
```diff
@@ -0,0 +1,77 @@
...
+import { fetch, FetchResultTypes } from '@utils/util';
+
+/**
+ * This class will also be used for handling complete games of Trivia
+ */
+// eslint-disable-next-line @typescript-eslint/no-extraneous-class
```
kyranet: This rule is enforced in Skyra and should not be disabled. Static classes are similar to namespaces, and as such, just using `export async function getQuestion(...) { ... }` is prefered.

src/lib/util/Games/TriviaManager.ts
```diff
@@ -0,0 +1,77 @@
...
+export const enum QuestionType {
+	BOOLEAN = 'boolean',
+	MULTIPLE = 'multiple'
+}
+
+export const enum QuestionDifficulty {
+	EASY = 'easy',
+	MEDIUM = 'medium',
+	HARD = 'hard'
+}
```
kyranet: Our convention on enum member names are the same as C#'s: `PascalCase`.

kyranet (changes requested):

src/commands/Games/trivia.ts
Stitch07: I don't think we should do this: the API doesn't have i18n support so it will always return True or False. Plus multiple-choice questions will always be in English (for aforementioned reasons), so I'd prefer boolean questions be in English, for consistency.

kyranet (approved):

favna (approved):

PR #1182, opened 2020-08-03
refactor(suggestion): follow normal suggestion flow for global suggestions

### LINKED ISSUE #1026: refactor: `suggest --global` should add suggestions to skyra-lounge configured suggestions
Right now `suggest --global` just calls the feedback command to revert to the legacy behaviour of `suggest` being aliased to `feedback`. The more ideal setup would be that `suggest --global` would go through the normal suggest flow, however the "guild" would be forced to skyra-lounge and create a suggestion in there.

This is separate from #777 in order to not overflow the scope of that PR.

Stitch07:
resolves https://github.com/skyra-project/skyra/issues/1026

### DIFF
--- a/src/commands/Suggestions/suggest.ts
+++ b/src/commands/Suggestions/suggest.ts
@@ -2,9 +2,10 @@ import { DbSet } from '@lib/structures/DbSet';
 import { SkyraCommand, SkyraCommandOptions } from '@lib/structures/SkyraCommand';
 import { PermissionLevels } from '@lib/types/Enums';
 import { GuildSettings } from '@lib/types/settings/GuildSettings';
+import { SUPPORT_SERVER } from '@root/config';
 import { ApplyOptions } from '@skyra/decorators';
 import { BrandingColors } from '@utils/constants';
-import { BitFieldResolvable, MessageEmbed, PermissionString, TextChannel } from 'discord.js';
+import { BitFieldResolvable, MessageEmbed, PermissionString, TextChannel, Webhook } from 'discord.js';
 import type { KlasaMessage } from 'klasa';
 
 const requiredChannelPermissions = ['SEND_MESSAGES', 'READ_MESSAGE_HISTORY', 'VIEW_CHANNEL'] as BitFieldResolvable<PermissionString>;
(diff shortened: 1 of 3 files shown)


src/commands/Suggestions/suggest.ts
```diff
@@ -24,17 +25,22 @@ export default class extends SkyraCommand {
 	private kChannelPrompt = this.definePrompt('<channel:textChannel>');
 
 	public async run(message: KlasaMessage, [suggestion]: [string]) {
-		// If including a flag of `--global` then revert to legacy behaviour of sending feedback
-		if (Reflect.has(message.flagArgs, 'global')) {
-			return this.client.commands.get('feedback')!.run(message, [suggestion]);
+		// If including a flag of `--global` send suggestions to #feedbacks in Skyra Lounge
+		const globalSuggestion = Reflect.has(message.flagArgs, 'global') && this.client.webhookFeedback;
+
+		const guild = globalSuggestion ? this.client.guilds.get(SUPPORT_SERVER)! : message.guild!;
```
kyranet: This is a bold asumption. One of the reasons why we use a webhook is because they're available anytime regardless of state **and** shard.

If Skyra starts sharding, this code will not longer work and will cause `TypeError`s to be thrown.

kyranet (dismissed):

src/commands/Suggestions/suggest.ts
Favna: yeah my idea was that the suggestion would be send by the webhook

src/config.example.ts
```diff
@@ -16,6 +16,7 @@ export const WSS_PORT = 565;
 
 export const NAME = 'Skyra';
 export const PREFIX = 'sd!';
+export const SUPPORT_SERVER = '254360814063058944';
```
favna: We can use `WEBHOOK_FEEDBACK.guild_id` for this btw, I just realised.

src/commands/Suggestions/suggest.ts
favna: Added a note about it and since we need to fix more for sharding that'll have to do for now

favna (approved):

PR #1183, opened 2020-08-04
patch(createMute): react to original message

Stitch07:
Currently, users have no way of knowing createMute with an existing role actually worked.

### DIFF
--- a/src/commands/Management/createMute.ts
+++ b/src/commands/Management/createMute.ts
@@ -1,6 +1,9 @@
 import { SkyraCommand } from '@lib/structures/SkyraCommand';
 import { PermissionLevels } from '@lib/types/Enums';
 import { GuildSettings } from '@lib/types/settings/GuildSettings';
+import { Emojis } from '@utils/constants';
+import { displayEntry } from '@utils/SettingsUtils';
+import { resolveEmoji } from '@utils/util';
 import { Role } from 'discord.js';
 import { CommandStore, KlasaMessage } from 'klasa';
 
@@ -24,9 +27,12 @@ export default class extends SkyraCommand {
 			const [role] = (await this.rolePrompt
 				.createPrompt(message, { time: 30000, limit: 1 })
 				.run(message.language.tget('ACTION_SHARED_ROLE_SETUP_EXISTING_NAME'))) as [Role];
-			await message.guild!.settings.update(GuildSettings.Roles.Muted, role, {
+
+			const [update] = await message.guild!.settings.update(GuildSettings.Roles.Muted, role, {
 				extraContext: { author: message.author.id }
 			});
+			if (message.reactable) return message.react(resolveEmoji(Emojis.GreenTick)!);
+			return message.sendLocale('COMMAND_CONF_UPDATED', [GuildSettings.Roles.Muted, displayEntry(update.entry, update.next, message.guild!)]);
 		} else if (await message.ask(message.language.tget('ACTION_SHARED_ROLE_SETUP_NEW'))) {
 			await message.guild!.security.actions.muteSetup(message);
 			await message.sendLocale('COMMAND_SUCCESS');


src/commands/Management/createMute.ts
```diff
@@ -26,6 +28,7 @@ export default class extends SkyraCommand {
 			await message.guild!.settings.update(GuildSettings.Roles.Muted, role, {
 				extraContext: { author: message.author.id }
 			});
+			await message.react(resolveEmoji(Emojis.GreenTick)!);
```
PyroTechniac: I don't think you have to resolve the emoji, as GreenTick is already a valid emoji

PyroTechniac (dismissed):

src/commands/Management/createMute.ts
kyranet: For texts? Yes.
For reacts? No.

Stitch07: [EmojiResolvables](https://discord.js.org/#/docs/main/stable/typedef/EmojiIdentifierResolvable) need IDs or Emoji objects for custom emotes.

vladfrangu: Shouldn't this report the same thing as when you create?

Aka

```ts
await message.sendLocale('COMMAND_SUCCESS');
```

vladfrangu (changes requested):

src/commands/Management/createMute.ts
Stitch07: I prefer reacting instead of stating the obvious (using the role "Muted"), but it's @kyranet's bot so I'll let him decide.

favna (dismissed):

src/commands/Management/createMute.ts
kyranet: The permissions are not checked, you're taking for granted that Skyra will have all `ADD_REACTIONS`, `USE_EXTERNAL_EMOJIS`, and `READ_CHANNEL_HISTORY` permissions, while it's rare for her to lack any of those, it can happen.

kyranet (changes requested):

src/commands/Management/createMute.ts
Stitch07: I've added the extra permissions to `requiredGuildPermissions`.

src/commands/Management/createMute.ts
```diff
@@ -15,7 +17,7 @@ export default class extends SkyraCommand {
 			description: language => language.tget('COMMAND_CREATEMUTE_DESCRIPTION'),
 			extendedHelp: language => language.tget('COMMAND_CREATEMUTE_EXTENDED'),
 			permissionLevel: PermissionLevels.Administrator,
-			requiredGuildPermissions: ['MANAGE_CHANNELS', 'MANAGE_ROLES'],
+			requiredGuildPermissions: ['MANAGE_CHANNELS', 'MANAGE_ROLES', 'USE_EXTERNAL_EMOJIS', 'ADD_REACTIONS', 'READ_MESSAGE_HISTORY'],
```
kyranet: Those permissions are guild-wide, Skyra might have them revoked specifically in the channel, or granted.

What I meant was to add `if (message.reactable) await message.react(...);`, maybe with a text branch as fallback.

kyranet (changes requested):

src/commands/Management/createMute.ts
Stitch07: Done

kyranet (dismissed):

src/commands/Management/createMute.ts
```diff
@@ -15,17 +18,20 @@ export default class extends SkyraCommand {
...
```
vladfrangu: I am against this. This will make the command unusable if Skyra doesn't have these permissions guild wise, which it doesn't even NEED for updating a configuration key. Imo, remove this and the react thing. Consistency please

vladfrangu (dismissed):

src/commands/Management/createMute.ts
Stitch07: I've removed the extra permissions, but am keeping the react (with a fallback to text).

vladfrangu: That's *fine*, but it still creates a massive inconsistency between command outputs :/

favna (approved):

PR #1186, opened 2020-08-06
feat: country command

Stitch07:


favna (dismissed):

favna (dismissed):

src/commands/Tools/Websearch/country.ts
```diff
@@ -0,0 +1,103 @@
...
+		await display.start(response, message.author.id);
+		return response;
+	}
+
+	private async fetchAPI(message: KlasaMessage, countryName: string) {
+		const apiResult = await fetch<CountryApiResult>(`https://restcountries.eu/rest/v2/name/${encodeURIComponent(countryName)}`)
+			.catch(() => { throw message.language.tget('SYSTEM_QUERY_FAIL'); });
+		if (Reflect.has(apiResult, 'status') && (apiResult as CountryResultError).status === 404) {
+			throw message.language.tget('SYSTEM_QUERY_FAIL');
+		}
```
kyranet: Does the API return 200 OK on a 404? `utils.fetch` throws on 400-599 status codes.

kyranet (changes requested):

src/commands/Tools/Websearch/country.ts
Stitch07: It returns 404 NOT FOUND, I'll remove the check.

kyranet (approved):

kyranet (approved):

PR #1193, opened 2020-08-10
refactor(avatar): flag for size

Stitch07:


src/commands/Tools/avatar.ts
```diff
@@ -1,10 +1,14 @@
...
 import { SkyraCommand } from '@lib/structures/SkyraCommand';
-import { MessageEmbed } from 'discord.js';
+import { MessageEmbed, ImageExt, ImageSize } from 'discord.js';
 import { CommandStore, KlasaMessage, KlasaUser } from 'klasa';
 
+const VALID_SIZES = [32, 64, 128, 256, 512, 1024, 2048];
+
 export default class extends SkyraCommand {
 
+	private kExtensionRegex = /.?(png|jpg|webp|gif)/i;
```
vladfrangu: Escape the dot and allow jpeg
```suggestion
	private kExtensionRegex = /\.?(png|jpe?g|webp|gif)/i;
```

src/commands/Tools/avatar.ts
```diff
@@ -13,20 +17,37 @@ export default class extends SkyraCommand {
...
+			.setAuthor(user.tag, user.displayAvatarURL({ size: 128, format: ext, dynamic: true }))
 			.setColor(await DbSet.fetchColor(message))
-			.setImage(user.displayAvatarURL({ size: 2048, format: 'png', dynamic: true })));
+			.setImage(user.displayAvatarURL({ size, format: ext, dynamic: true })));
+	}
+
+	private resolveExt(ext: string): ImageExt {
+		const results = this.kExtensionRegex.exec(ext);
+		this.kExtensionRegex.lastIndex = 0;
+		if (results) return results[0].toLowerCase() as ImageExt;
```
vladfrangu: For a string like `png`, this is fine, but for one like `.png`, it isn't 
image

```suggestion
		if (results) return results[1].toLowerCase() as ImageExt;
```

vladfrangu (dismissed):

src/commands/Tools/avatar.ts
Stitch07: I've talked to Favna and I'm removing the extension flag entirely, so this will not matter.

kyranet (dismissed):
`--ext` was removed but is still referenced in the i18n strings.

favna (approved):

PR #1225, opened 2020-08-21
fix(dehoist): question usernames as well

cfanoulis:
Fixes recent issue popping up in Discord

Tested, needs reviews please
Target: weekend release

### DIFF
--- a/src/commands/Moderation/dehoist.ts
+++ b/src/commands/Moderation/dehoist.ts
@@ -28,7 +28,7 @@ export default class extends SkyraCommand {
 
 		for (const [memberId, memberTag] of members.manageableMembers()) {
 			const displayName = memberTag.nickname ?? this.client.userTags.get(memberId)!.username;
-			if (!displayName) return;
+			if (!displayName) continue;
 
 			const char = displayName.codePointAt(0)!;
 			if (char > this.kLowestCode) continue;
--- a/src/commands/Moderation/dehoist.ts
+++ b/src/commands/Moderation/dehoist.ts
@@ -27,11 +27,14 @@ export default class extends SkyraCommand {
 		const response = await message.sendLocale('SYSTEM_LOADING', []);
 
 		for (const [memberId, memberTag] of members.manageableMembers()) {
-			const displayName = memberTag.nickname ?? this.client.users.get(memberId)?.username;
+			const displayName = memberTag.nickname ?? this.client.userTags.get(memberId)!.username;
+			if (!displayName) return;
+			const char = displayName.codePointAt(0)!;
 
-			if (displayName && displayName.codePointAt(0)! < this.kLowestCode) {
+			if (char < this.kLowestCode) {
 				// Replace the first character of the offending user's with a downwards arrow, bringing'em down, down ,down
-				const newNick = `🠷 ${displayName.slice(2)}`;
+				// The ternary cuts 2 characters if the 1st codepoint belongs in UTF-16
+				const newNick = `🠷${displayName.slice(char <= 0xff ? 1 : 2)}`;
 				try {
 					await api(this.client)
 						.guilds(message.guild!.id)
@@ -46,12 +49,12 @@ export default class extends SkyraCommand {
 
 		// We're done!
 		return response.edit({
-			embed: await this.prepareFinalEmbed(message, members.size, counter, errored),
+			embed: await this.prepareFinalEmbed(message, counter, errored),
 			content: null
 		});
 	}
 
-	private async prepareFinalEmbed(message: KlasaMessage, totalMembers: number, dehoistedMembers: number, erroredChanges: ErroredChange[]) {
+	private async prepareFinalEmbed(message: KlasaMessage, dehoistedMembers: number, erroredChanges: ErroredChange[]) {
 		const embedLanguage = message.language.get('COMMAND_DEHOIST_EMBED', {
 			dehoistedMemberCount: dehoistedMembers,
 			dehoistedWithErrorsCount: dehoistedMembers - erroredChanges.length,
--- a/src/commands/Moderation/dehoist.ts
+++ b/src/commands/Moderation/dehoist.ts
@@ -31,9 +31,9 @@ export default class extends SkyraCommand {
 			if (!displayName) return;
 
 			const char = displayName.codePointAt(0)!;
-
 			if (char > this.kLowestCode) continue;
-			// Replace the first character of the offending user's with a downwards arrow, bringing'em down, down ,down
+
+			// Replace the first character of the offending user's with an UTF-16 character, bringing'em down, down, down.
 			// The ternary cuts 2 characters if the 1st codepoint belongs in UTF-16
 			const newNick = `🠷${displayName.slice(char <= 0xff ? 1 : 2)}`;
 			try {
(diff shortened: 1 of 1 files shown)


src/commands/Moderation/dehoist.ts
```diff
@@ -27,16 +27,18 @@ export default class extends SkyraCommand {
 		const response = await message.sendLocale('SYSTEM_LOADING', []);
 
 		for (const [memberId, memberTag] of members.manageableMembers()) {
-			if (memberTag.nickname && memberTag.nickname.codePointAt(0)! < this.kLowestCode) {
+			const displayName = memberTag.nickname ?? this.client.users.get(memberId)?.username;
+
+			if (displayName && displayName.codePointAt(0)! < this.kLowestCode) {
 				// Replace the first character of the offending user's with a downwards arrow, bringing'em down, down ,down
-				const newNick = `🠷${memberTag.nickname.slice(1)}`;
+				const newNick = `🠷 ${displayName.slice(2)}`;
```
kyranet: But... why remove two? If anything what you should do is:

1. Continue early if `displayName` is undefined.
1. Store `displayName.codePointAt(0)` as a constant.
1. Check if the constant is lower than `kLowestCode`.
1. If it is, check the byte range, `codePoint <= 0xFF ? 1 : 2`, assign it to a constant.
1. Call `.slice()` with the previous constant.

🎉! UTF-16 support!

kyranet (changes requested):

cfanoulis:
I remove 2 because I use the special char + a space, for visuals

uh I guess I'll do what you did about UTF-18 tho

kyranet:
Most of the symbols used to hoist are "noisy" in a sense, e.g. `!`. For visuals, people add themselves a space. However, this is not always the case. If I set my nickname as `!kyra`, this would change it to `🠷 yra`.

src/commands/Moderation/dehoist.ts
cfanoulis: resolved in 52b0a266b8949

src/commands/Moderation/dehoist.ts
```diff
@@ -27,29 +27,34 @@ export default class extends SkyraCommand {
 		const response = await message.sendLocale('SYSTEM_LOADING', []);
 
 		for (const [memberId, memberTag] of members.manageableMembers()) {
-			if (memberTag.nickname && memberTag.nickname.codePointAt(0)! < this.kLowestCode) {
+			const displayName = memberTag.nickname ?? this.client.userTags.get(memberId)!.username;
+			if (!displayName) return;
+			const char = displayName.codePointAt(0)!;
+
```
kyranet:
```suggestion
			const displayName = memberTag.nickname ?? this.client.userTags.get(memberId)!.username;
			if (!displayName) continue;

			const char = displayName.codePointAt(0)!;
```

src/commands/Moderation/dehoist.ts
```diff
@@ -27,29 +27,34 @@ export default class extends SkyraCommand {
...
 					await api(this.client)
 						.guilds(message.guild!.id)
 						.members(memberId)
 						.patch({ data: { nick: newNick }, reason: 'Dehoisting' });
 				} catch (error) {
-					errored.push({ oldNick: memberTag.nickname, newNick });
+					errored.push({ oldNick: displayName, newNick });
 				}
 				counter++;
 			}
```
kyranet: Flatten this with `if (char >= this.kLowestCode) continue;`

kyranet (changes requested):

src/commands/Moderation/dehoist.ts
cfanoulis: resolved in f6c57c8

src/commands/Moderation/dehoist.ts
cfanoulis: resolved in f6c57c8

src/commands/Moderation/dehoist.ts
```diff
@@ -27,29 +27,34 @@ export default class extends SkyraCommand {
...
-				} catch (error) {
-					errored.push({ oldNick: memberTag.nickname, newNick });
-				}
-				counter++;
+			const displayName = memberTag.nickname ?? this.client.userTags.get(memberId)!.username;
+			if (!displayName) return;
+
+			const char = displayName.codePointAt(0)!;
+
+			if (char > this.kLowestCode) continue;
```
kyranet:
```suggestion
			const char = displayName.codePointAt(0)!;
			if (char > this.kLowestCode) continue;

```

src/commands/Moderation/dehoist.ts
```diff
@@ -27,29 +27,34 @@ export default class extends SkyraCommand {
...
+			// Replace the first character of the offending user's with a downwards arrow, bringing'em down, down ,down
```
kyranet:
```suggestion
			// Replace the first character of the offending user's with an UTF-16 character, bringing'em down, down, down.
```

src/commands/Moderation/dehoist.ts
```diff
@@ -27,29 +27,34 @@ export default class extends SkyraCommand {
...
+			// The ternary cuts 2 characters if the 1st codepoint belongs in UTF-16
+			const newNick = `🠷${displayName.slice(char <= 0xff ? 1 : 2)}`;
+			try {
+				await api(this.client)
+					.guilds(message.guild!.id)
+					.members(memberId)
+					.patch({ data: { nick: newNick }, reason: 'Dehoisting' });
+			} catch (error) {
```
kyranet:
```suggestion
			} catch {
```

src/commands/Moderation/dehoist.ts
```diff
@@ -27,29 +27,34 @@ export default class extends SkyraCommand {
...
+				errored.push({ oldNick: displayName, newNick });
 			}
+			counter++;
```
kyranet:
```suggestion

			++counter;
```

kyranet (changes requested):

src/commands/Moderation/dehoist.ts
cfanoulis: does it make a difference? we don't access the value either way, unless one is more performant or something than the other?

kyranet: More performant as it doesn't clone the value but just increases it. It's also what we do everywhere else in Skyra.
I also added one line of padding for better readability.

src/commands/Moderation/dehoist.ts
```diff
@@ -27,29 +27,35 @@ export default class extends SkyraCommand {
...
-					await api(this.client)
-						.guilds(message.guild!.id)
-						.members(memberId)
-						.patch({ data: { nick: newNick }, reason: 'Dehoisting' });
-				} catch (error) {
-					errored.push({ oldNick: memberTag.nickname, newNick });
-				}
-				counter++;
+			const displayName = memberTag.nickname ?? this.client.userTags.get(memberId)!.username;
+			if (!displayName) return;
```
kyranet:
```suggestion
			if (!displayName) continue;
```

kyranet (changes requested):

kyranet (approved):

cfanoulis:
My dumbass forgot and deleted the repo, moved to #1228

PR #1226, opened 2020-08-22
refactor: migrate to @sapphire/utilities

favna:
- refactor(*): migrate to @sapphire/utilities
- style(*): formatting
- refactor(skyraclient): replace last @klasa/utils reference


kyranet (commented):
Coverage needs to be fixed.

package.json
```diff
@@ -44,8 +44,8 @@
 		"url": "https://github.com/skyra-project/skyra/issues"
 	},
 	"dependencies": {
-		"@influxdata/influxdb-client": "^1.6.0",
 		"@influxdata/influxdb-client-apis": "^1.6.0",
+		"@influxdata/influxdb-client": "^1.6.0",
```
Quantumlyy, line 48: Why was this moved?

kyranet: Because the sort utility we use from VSCode detects `"` as a character of a higher code than `-`, so it puts the entry below, while yarn sorts by the names (which would place this one above).

That's my guess.

favna: yeah that indeed. It's just VSC's standard "sort alphabetically". After a while yarn will probably shift it back but I cbb to do it manually.

Quantumlyy:
> Coverage needs to be fixed.

Couldn't we maybe do tests for something else to fix coverage that way?

kyranet (approved):

PR #1228, opened 2020-08-22
fix(dehoist): question usernames as well

cfanoulis:
dumbass energy goes brr

### DIFF
--- a/src/commands/Moderation/dehoist.ts
+++ b/src/commands/Moderation/dehoist.ts
@@ -28,7 +28,7 @@ export default class extends SkyraCommand {
 
 		for (const [memberId, memberTag] of members.manageableMembers()) {
 			const displayName = memberTag.nickname ?? this.client.userTags.get(memberId)!.username;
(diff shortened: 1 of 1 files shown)


src/commands/Moderation/dehoist.ts
```diff
@@ -27,29 +27,35 @@ export default class extends SkyraCommand {
...
-					await api(this.client)
-						.guilds(message.guild!.id)
-						.members(memberId)
-						.patch({ data: { nick: newNick }, reason: 'Dehoisting' });
-				} catch (error) {
-					errored.push({ oldNick: memberTag.nickname, newNick });
-				}
-				counter++;
+			const displayName = memberTag.nickname ?? this.client.userTags.get(memberId)!.username;
+			if (!displayName) return;
```
kyranet:
```suggestion
			if (!displayName) continue;
```

kyranet (changes requested):

kyranet (approved):

PR #1239, opened 2020-08-24
fix: add proper ts support according to sentry

Stitch07:


src/Skyra.ts
```diff
@@ -4,16 +4,27 @@ import 'reflect-metadata';
...
-		sentry.init({ dsn: TOKENS.SENTRY_URL });
+		Sentry.init({
+			dsn: TOKENS.SENTRY_URL,
+			integrations: [
+				new Sentry.Integrations.Modules(),
+				new Sentry.Integrations.FunctionToString(),
+				new Sentry.Integrations.LinkedErrors(),
+				new Sentry.Integrations.Console(),
+				new Sentry.Integrations.Http({ breadcrumbs: true, tracing: true }),
+				new RewriteFrames({ root: __dirname || process.cwd() })
```
favna: we have a `rootDir` constant in `constants.ts`, better to use that one.

src/tsconfig.json
```diff
@@ -11,6 +11,8 @@
 			"@root/*": ["*"]
 		},
 		"composite": true,
+		"sourceMap": true,
+		"inlineSources": true,
```
favna:
```suggestion
```

`sourceMap` is inherited and `inlineSources` should not be needed. Sentry's page is pretty outdated for as far as I can tell tbh.

favna (changes requested):

favna (approved):

PR #1242, opened 2020-08-25
feat: poll enhancements

### LINKED ISSUE #1240: request: rework poll

The current `poll` command is more of a `vote` command. For a poll it would be good if it added numbers from 1 to 10, for each option given. Then rename the current `poll` to `vote`.

IF amount of options <= 10 THEN use numbers (with a min of 1)
IF amount of options > 10 THEN use letters (with a max of 20 as that's max reactions)

N.A.

**Additional context**

N.A.

Stitch07:
* Renamed `poll` to `vote`
* Added a simple poll command with reaction-based options

Closes #1240

### DIFF
--- a/src/commands/Tools/poll.ts
+++ b/src/commands/Tools/poll.ts
@@ -4,34 +4,7 @@ import { KlasaMessage } from 'klasa';
 
 const NUMBER_OPTS = ['1️⃣', '2️⃣', '3️⃣', '4️⃣', '5️⃣', '6️⃣', '7️⃣', '8️⃣', '9️⃣', '🔟'];
 
-const ALPHABET_OPTS = [
-	'🇦',
-	'🇧',
-	'🇨',
-	'🇩',
-	'🇪',
-	'🇫',
-	'🇬',
-	'🇭',
-	'🇮',
-	'🇯',
-	'🇰',
-	'🇱',
-	'🇲',
-	'🇳',
-	'🇴',
-	'🇵',
-	'🇶',
-	'🇷',
-	'🇸',
-	'🇹',
-	'🇺',
-	'🇻',
-	'🇼',
-	'🇽',
-	'🇾',
-	'🇿'
-];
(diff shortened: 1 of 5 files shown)


src/commands/Tools/poll.ts
```diff
@@ -2,20 +2,59 @@ import { SkyraCommand, SkyraCommandOptions } from '@lib/structures/SkyraCommand'
 import { ApplyOptions } from '@skyra/decorators';
 import { KlasaMessage } from 'klasa';
 
+const NUMBER_OPTS = ['1️⃣', '2️⃣', '3️⃣', '4️⃣', '5️⃣', '6️⃣', '7️⃣', '8️⃣', '9️⃣', '🔟'];
+
+const ALPHABET_OPTS = [
```
kyranet: This array is so obnoxiously long I'd like prettier to be ignored like in [char](https://github.com/skyra-project/char/blob/main/src/lib/CharUnicodeInfo/CategoriesValues.ts).

kyranet: Also, there are 25 elements, but the maximum is 20.

src/commands/Tools/poll.ts
```diff
@@ -2,20 +2,59 @@ import { SkyraCommand, SkyraCommandOptions } from '@lib/structures/SkyraCommand'
...
+
+		const emojis = (options.length > 10 ? ALPHABET_OPTS : NUMBER_OPTS).slice(0, options.length);
+		const loadingMsg = await message.sendLocale('systemLoading', []);
+
+		for (const emoji of emojis) {
+			await loadingMsg.react(emoji);
 		}
 
-		return message;
+		await message.send(options.map((option, i) => `${i + 1}. *${option.trim()}*`).join('\n'));
```
kyranet:
```suggestion
		await message.send(options.map((option, i) => `${emojis[i]} → *${option.trim()}*`).join('\n'));
```

kyranet (changes requested):

favna (approved):

kyranet (approved):

PR #1243, opened 2020-08-25
feat(dehoist): requested enhancements

### LINKED ISSUE #1241: request: dehoist enhancements

The dehoist command is pretty simple and doesn't cover an edge case besides it has a bad UX:

- The command makes no exception for `'0'..='9'` characters (which are lower codepoints compared to `'A'`.
- The command doesn't say how many users is it renaming.
- The command doesn't include the progress (ratelimits make dehoist slow for +10 members), instead, it keeps in loading state.

- [ ] Make exceptions for code points between `'0'` and `'9'`, inclusive.
- [ ] Edit the message with the number of hoisted members.
- [ ] Edit the message with the progress.
- [ ] (Optional) Rename while receiving packets from the gateway (and update accordingly), this allows Skyra to run dehoist on very large guilds.

N/a.

**Additional context**

N/a.

Stitch07:
Closes #1241 

> Rename while receiving packets from the gateway (and update accordingly), this allows Skyra to run dehoist on very large guilds.

I'm not sure what this is, but everything else has been added.


src/commands/Moderation/dehoist.ts
```diff
@@ -24,14 +28,17 @@ export default class extends SkyraCommand {
 		let counter = 0;
 		const errored: ErroredChange[] = [];
 		const members = message.guild!.memberTags;
-		const response = await message.sendLocale('systemLoading', []);
 
-		for (const [memberId, memberTag] of members.manageableMembers()) {
+		const manageableMembers = Array.from(members.manageableMembers());
+		const hoistedMembers = manageableMembers.filter(this.shouldDehoist.bind(this));
```
kyranet: Why the double iteration? I'm pretty sure you can do `members.manageableMembers().filter(this.shouldDehoist.bind(this))`.

src/commands/Moderation/dehoist.ts
```diff
@@ -55,16 +69,25 @@ export default class extends SkyraCommand {
...
 
+	private shouldDehoist([memberId, memberTag]: [string, MemberTag]) {
+		const displayName = memberTag.nickname ?? this.client.userTags.get(memberId)!.username;
+		if (!displayName) return false;
+
+		const char = displayName.codePointAt(0)!;
+
+		// number char codes are greater than kLowestCode, but still hoist the
+		if (char >= this.kLowestCode && !(char >= kLowestNumberCode && char <= kHighestNumberCode)) return false;
+		return true;
```
kyranet: You can negate the if and return it instead of adding extra conditionals.

src/lib/util/Cache/MemberTags.ts
```diff
@@ -88,6 +88,7 @@ export class MemberTags extends Collection<string, MemberTag> {
 		} else {
 			for (const tag of this.entries()) {
 				if (tag[1].roles.some((role) => nonManageableRoles.has(role))) continue;
+				if (tag[0] === this.kGuild.ownerID) continue;
```
kyranet, line 90: This call is far less expensive than the one above, please move it up.

src/commands/Moderation/dehoist.ts
```diff
@@ -46,6 +53,13 @@ export default class extends SkyraCommand {
 			}
 
 			++counter;
+
+			// update the counter every 10 dehoists
+			if ((i + 1) % 10 === 0) {
+				const dehoistPercentage = (i / hoistedMembers.length) * 100;
+				await message.sendLocale('commandDehoistProgress', [{ count: i + 1, percentage: Math.round(dehoistPercentage) }]);
+				await sleep(5000);
```
kyranet: Any reason for this random sleep? In a real world scenario, the message would be sent every 30 seconds or so, nickname changes are pretty ratelimited.

kyranet (dismissed):

src/commands/Moderation/dehoist.ts
Stitch07: I thought about doing that, but this way is a lot more readable imo.

favna (approved):

PR #1264, opened 2020-09-06
chore(deps): update discord.js to stable

### LINKED ISSUE #1244: request: switch to DiscordJS stable

In order to switch to Sapphire Framework we will need to upgrade to DiscordJS stable.

Among other things this means embracing managers.

Upgrade the version then fix all the broken code

N.A.
~~klasa/core lmfaono~~

**Additional context**

N.A.

Stitch07:
TODO:

* Finish klasa update: https://github.com/skyra-project/klasa/pull/1
* Actually test this version

Fixes #1244


kyranet:
Things we need to do for this to fully work:

- [x] Update SG (the `guilds` gateway should read `client.guilds.cache` instead of `client.guilds`), fix its detection system by adding an extra condition in the ternary.
- [x] Update Klasa for new SG and bugfixes.
- [x] Tweak Skyra's raw events, some of them were made to override discord.js's behaviour (presenceUpdate, guildCreate), but this is not longer possible as they removed `disabledEvents`. Instead, Skyra should use discord.js's regular events where possible (userUpdate, memberUpdate...), using raw only if it is to avoid the need of state (addMessageReaction, removeMessageReaction...) (which has been reduced in the last releases).

src/commands/Giveaway/greroll.ts
```diff
@@ -32,10 +32,10 @@ export default class extends SkyraCommand {
 		const content = winners
 			? message.language.get('giveawayEndedMessage', {
 					winners: winners.map((winner) => `<@${winner}>`),
-					title
+					title: title!
 			  })
-			: message.language.get('giveawayEndedMessageNoWinner', { title });
-		return message.sendMessage(content);
+			: message.language.get('giveawayEndedMessageNoWinner', { title: title! });
+		return message.sendMessage(content, { allowedMentions: { users: [message.author.id], roles: [] } });
```
favna: Self correction: winners should be mentionable

favna (changes requested):

src/events/guildCreate.ts
```diff
@@ -0,0 +1,34 @@
...
+
+		// Cache all tags and prepare for sweep
+		for (const member of guild.members.cache.values()) {
+			this.client.userTags.create(member.user);
+			guild.memberTags.create(member);
+
+			if (member.voice.channelID) membersInVoice.push(member);
+		}
+
+		const { me } = guild;
```
favna, line 16: @kyranet why do we have this anyway btw? Seems a bit useless considering there's only 1 usage of `guild.me` (like 29)

favna (changes requested):
Revert guildCreate, requesting this as a change to block merge

PR #1265, opened 2020-09-07
fix(UserRepository): check for gameName in gameIntegrations

adiologydev:
`fetchGameIntegrations` method in `UserRepository` would always return the first entry it would've found for a user in `gameIntegrations`, this will fix that so it now fetches it along with the specified `gameName`.


Stitch07 (approved):

kyranet (approved):

Stitch07:
@all-contributors please add @AdityaTD for bugs

PR #1292, opened 2020-09-19
feat(orm): added redis caching

kyranet:


favna (approved):

src/config.example.ts
```diff
@@ -28,6 +29,9 @@ export const PGSQL_DATABASE_PASSWORD = '';
 export const PGSQL_DATABASE_USER = '';
 export const PGSQL_DATABASE_PORT = 5432;
 export const PGSQL_DATABASE_HOST = 'localhost';
+export const PGSQL_CACHE_REDIS = false;
+
+export const REDIS_OPTIONS: RedisOptions = {};
```
favna: Fill this with port 8287 by default because that's where it runs on the vps for saelem. Also DB 0 is used by Saelem so we should default it to DB 1.

favna (changes requested):

favna (approved):

PR #1298, opened 2020-09-25
feat(weather): add --fahrenheit flag

Stitch07:


src/commands/Google/weather.ts
```diff
@@ -18,18 +19,23 @@ const COLORS = {
...
-		super(store, file, directory, {
-			bucket: 2,
-			cooldown: 120,
-			description: (language) => language.get('commandWeatherDescription'),
-			extendedHelp: (language) => language.get('commandWeatherExtended'),
-			requiredPermissions: ['ATTACH_FILES'],
-			usage: '<city:string>'
-		});
-	}
+const farenheitToCelsius = (celsius: number) => celsius * 1.8 + 32;
```
kyranet: This... is converting Celsius to Fahrenheit.

src/commands/Google/weather.ts
```diff
@@ -18,18 +19,23 @@ const COLORS = {
...
+
+const enum TemperatureUnit {
+	Celsius,
+	Farenheit
```
kyranet:
```suggestion
	Fahrenheit
```

src/languages/en-US.ts
```diff
@@ -2742,7 +2742,7 @@ export default class extends Language {
 			],
 			explainedUsage: [['city', 'The locality, governing, country or continent to check the weather from.']],
 			examples: ['Antarctica', 'Arizona'],
-			reminder: 'Temperature is in **Celsius**',
+			reminder: 'Temperature is in **Celsius** by default. Use the --imperial or --farenheit flag to view it in **Farenheit**.',
```
kyranet:
```suggestion
			reminder: 'Temperature is in **Celsius** by default. Use the --imperial or --fahrenheit flag to view it in **Farenheit**.',
```

src/commands/Google/weather.ts
```diff
@@ -39,6 +45,8 @@ export default class extends SkyraCommand {
 		let country = '';
 		let continent = '';
 
+		const useImperial = Reflect.has(message.flagArgs, 'farenheit') || Reflect.has(message.flagArgs, 'imperial');
```
kyranet:
```suggestion
		const useImperial = Reflect.has(message.flagArgs, 'fahrenheit') || Reflect.has(message.flagArgs, 'imperial');
```

src/languages/es-ES.ts
```diff
@@ -2776,7 +2776,8 @@ export default class extends Language {
 			],
 			explainedUsage: [['ciudad', 'La localidad, el gobierno, el país o el continente para consultar la hora.']],
 			examples: ['Madrid', 'Barcelona'],
-			reminder: 'La temperatura está en **Celsius**',
+			reminder:
+				'La temperatura está en ** Celsius ** de forma predeterminada. Use la bandera --imperial o --farenheit para verla en ** Farenheit **.',
```
kyranet:
```suggestion
				'La temperatura está en ** Celsius ** de forma predeterminada. Use la bandera --imperial o --fahrenheit para verla en ** Farenheit **.',
```

kyranet (changes requested):

src/commands/Google/weather.ts
```diff
@@ -18,18 +19,23 @@ const COLORS = {
...
-		super(store, file, directory, {
-			bucket: 2,
-			cooldown: 120,
-			description: (language) => language.get('commandWeatherDescription'),
-			extendedHelp: (language) => language.get('commandWeatherExtended'),
-			requiredPermissions: ['ATTACH_FILES'],
-			usage: '<city:string>'
-		});
-	}
+const celsiusToFahrenheit = (celsius: number) => celsius * 1.8 + 32;
```
favna:
```suggestion
const celsiusToFahrenheit = (celsius: number) => (celsius * 9 / 5) + 32;
```

9 / 5 is 1.8 yes, but the official formula has `9 / 5` which is the official conversion (and because 5 / 9 for the other way around is not roundable (0.55555555555555555555556 basically))

image

favna (changes requested):

src/commands/Google/weather.ts
Stitch07: While deriving the actual formula, you multiply it by `180 / 100` which is simplified to `9/5`. So 1.8 should be fine.
[Source](https://www.mathsisfun.com/temperature-conversion.html)

kyranet: We prioritise precision over speed, which is why our saturation calculation is the complex variant.

favna (approved):

kyranet (approved):

PR #1302, opened 2020-09-28
fix: donate command

Stitch07:


src/commands/System/donate.ts
```diff
@@ -14,10 +14,11 @@ export default class extends SkyraCommand {
 	public async run(message: KlasaMessage) {
 		try {
 			// TODO(Soumil07): Extract i18n help formatter so it's useable here
```
favna: You can remove this TODO if it's fixed. It was in place here to fix donate.

favna (changes requested):

kyranet (approved):

favna (approved):

PR #1303, opened 2020-09-30
fix(weather): fixed celsius to fahrenheit conversion formula

kyranet:


src/commands/Google/weather.ts
```diff
@@ -20,7 +20,7 @@ const COLORS = {
 	windy: '#33B679'
 };
 
-const celsiusToFahrenheit = (celsius: number) => (celsius * 5) / 9 + 32;
+const celsiusToFahrenheit = (celsius: number) => celsius * (9 / 5) + 32;
```
favna:
```suggestion
const celsiusToFahrenheit = (celsius: number) => (celsius * 9 / 5) + 32;
```

image


Same what I showed Stitch at the time

favna (changes requested):

src/commands/Google/weather.ts
```diff
@@ -20,7 +20,7 @@ const COLORS = {
 	windy: '#33B679'
 };
 
-const celsiusToFahrenheit = (celsius: number) => (celsius * 5) / 9 + 32;
+const celsiusToFahrenheit = (celsius: number) => (celsius * 9 / 5) + 32;
```
favna:
```suggestion
const celsiusToFahrenheit = (celsius: number) => (celsius * 9) / 5 + 32;
```

favna (changes requested):

favna (approved):

PR #1309, opened 2020-09-30
feat(disabled): add link to support server in globally disabled commands messages

favna:


src/languages/en-US.ts
```diff
@@ -217,7 +217,8 @@ export default class extends Language {
 		inhibitorRunin: ({ type }) => `This command is only available in ${type} channels.`,
 		inhibitorRuninNone: ({ name }) => `The ${name} command is not configured to run in any channel.`,
 		inhibitorDisabledGuild: 'This command has been disabled by an admin in this guild!',
-		inhibitorDisabledGlobal: 'This command has been globally disabled by the bot owners!',
+		inhibitorDisabledGlobal:
+			'This command has been globally disabled by the bot owners. Want to know why and find out when it will be back? Join the official Skyra server: https://join.skyra.pw',
```
kyranet:
```suggestion
			'This command has been globally disabled by the bot owners. Want to know why and find out when it will be back? Join the official Skyra server: <https://join.skyra.pw>',
```

kyranet: Also the message feels like downright advertising.

kyranet (changes requested):

src/languages/en-US.ts
favna: Advertising to our own users... how is that bad 🤔

kyranet: It just feels... a little aggresive.

favna: Do you want Skyra to grow or not... In the grand scheme of things this is probably one of the least aggressive steps we have to take to ever rise to the size of Pengu.

favna: Note that I don't disagree with you, contrary, I agree. But I also think it's acceptable. Especially considering we don't disable commands that often and this only applies to commands that *we* disabled, not commands disabled in a specific server

favna: For the record the error is thrown here: https://github.com/skyra-project/klasa/blob/settings/src/inhibitors/disabled.js

kyranet (dismissed):

kyranet (approved):

PR #1313, opened 2020-09-30
refactor: replace key in with reflect has

favna:


src/events/rawMessageReactionAdd.ts
```diff
@@ -30,7 +30,7 @@ export default class extends Event {
 			emoji: {
 				animated: raw.emoji.animated ?? false,
 				id: raw.emoji.id,
-				managed: 'managed' in raw.emoji ? raw.emoji.managed! : null,
+				managed: Reflect.get(raw.emoji, 'managed') || null,
```
kyranet: Get to 2020.
```suggestion
				managed: Reflect.get(raw.emoji, 'managed') ?? null,
```

kyranet (dismissed):

src/events/rawMessageReactionAdd.ts
favna: I saw the type was undefined or boolean so I thought maybe it has to catch off `false` and transform it to `null` as well :\

kyranet (approved):

PR #1348, opened 2020-10-20
feat(weeb): added better error handling for weeb.sh

kyranet:
We sometimes get 403 Unauthorized, and sometimes we get 5XX error codes.
Users get confused and keep trying even thought it doesn't work, so I added special messages.

### DIFF
--- a/src/languages/es-ES.ts
+++ b/src/languages/es-ES.ts
@@ -2900,6 +2900,8 @@ export default class extends Language {
 			explainedUsage: [['user', 'The user to tickle.']],
 			examples: ['@Skyra']
 		},
+		commandWeebUnavailableError: 'I am sorry, but the source of all my weeb images is not available at the moment. Please try again later!',
+		commandWeebUnexpectedError: 'Weird! I encountered an error I should have never received!',
 		commandWbangDescription: 'Bang 💥🔫!',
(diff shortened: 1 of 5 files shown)


Stitch07 (approved):

src/languages/es-ES.ts
```diff
@@ -2900,6 +2900,8 @@ export default class extends Language {
 			explainedUsage: [['user', 'The user to tickle.']],
 			examples: ['@Skyra']
 		},
+		commandWeebUnavailableError: 'I am sorry, but the source of all my weeb images is not available at the moment. Please try again later!',
+		commandWeebUnexpectedError: 'Weird! I encountered an error I should have never received!',
```
cfanoulis, line 2904: allow me to imitate favna:
*khm khm*

rEEEE ENGLISH IN THE SPANISH FILES

cfanoulis (commented):
hello i am enkiel and i have a nit

favna (approved):

src/languages/es-ES.ts
favna: Much easier to do post-i18n coz then we can remove all english keys in Spanish file as it'll fall back automatically

PR #1352, opened 2020-10-21
refactor(audio): allow no redis

favna:
This was previously coded to always try to connect to Redis, even when
Lavalink was disabled. This causes issues in development.

`!.` is safe for events because they are never emitted if Lavalink is
off

`!.` is save for `klasaReady` and `SkyraGuild` because those are not
accessed when Lavalink is off

`?.`  plus ?? is required for Analytics because Influx expects the
function to return points in integer.

### DIFF
--- a/src/lib/audio/QueueClient.ts
+++ b/src/lib/audio/QueueClient.ts
@@ -1,3 +1,4 @@
+import { ENABLE_LAVALINK } from '@root/config';
 import { Node, NodeOptions, NodeSend } from '@skyra/audio';
 import * as Redis from 'ioredis';
 import { QueueStore } from './QueueStore';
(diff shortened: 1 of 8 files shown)


src/lib/audio/QueueClient.ts
```diff
@@ -7,10 +8,12 @@ export interface QueueClientOptions extends NodeOptions {
 }
 
 export class QueueClient extends Node {
-	public readonly queues: QueueStore;
+	public readonly queues?: QueueStore;
 
 	public constructor(options: QueueClientOptions, send: NodeSend) {
 		super(options, send);
-		this.queues = new QueueStore(this, options.redis instanceof Redis ? options.redis : new Redis(options.redis));
+		if (ENABLE_LAVALINK) {
```
kyranet, line 15: The queue should be created on connect, to be honest, this will require some moves in the config files.

kyranet (changes requested):

Quantumlyy (approved):

src/lib/audio/QueueClient.ts
favna: that's out of scope for this PR.

kyranet: It still is, because it fixes the issue about Skyra always requiring Redis.

favna: well in that case I have no idea what needs to be done coz I haven't worked with Skyra Audio before soooo ¯\\\_(ツ)_/¯

kyranet: I'll get around this later, please create an issue regarding this so I can remember.

kyranet (approved):

PR #1301, opened 2020-09-28
refactor: remove SettingsGateway

kyranet:
Fixes #1279.

- [x] Fill metadata for all keys in GuildEntity.
- [x] Fill metadata for complex keys (updatable via dashboard, soon also by Sapphire's dynamic argument system).
- [x] Make system to validate data.
- [x] Remove everything SettingsGateway related.
- [x] Add extendables for easy control of (guild) settings.
- [x] Update Klasa fork to remove all traces of SettingsGateway.
- [x] Update Skyra to patched commit of fork.
- [x] Write migration for `TEXT[]` to `JSONB`.
- [x] Merge Klasa PR (https://github.com/skyra-project/klasa/pull/3) then change Klasa dependency back to settings branch
- [x] Merge skyra.pw PR (https://github.com/skyra-project/skyra.pw/pull/198)
- [x] Reset depfu to target `main` branch for both this repo and skyra.pw
- [x] Test everything.
- [x] $$Profit$$

Check to do before merging:
- [x] formatting all
- [x] `readSettings` using array where possible
- [x] using `settings.getLanguage()` where possible
- [x] using `GuildMessage` when `runIn` is `['text']`
- [x] removing non-null assertion in `message.guild!.` when `runIn` is `['text']`
- [x] replacing `const { ... } = message.guild.readSettings((settings) => ({ ... }));` with `const [ ... ] = message.guild.readSettings((settings) => [ ... ])`
- [x] replacing `const { ... } = message.guild.writeSettings((settings) => ({ ... }));` with `const [ ... ] = message.guild.writeSettings((settings) => [ ... ])`
- [x] replacing `settings.key` with `settings[GuildSettings.PathToKey]`
- [x] temporarily enable `@typescript-eslint/no-unnecessary-type-assertion` as `"@typescript-eslint/no-unnecessary-type-assertion": "warn"` to check for this rule

PR #1353, opened 2020-10-21
fix(serializer): allow news and store

favna:
fix this until #1301 is done

I have no idea what the proper type for `value` is.

Priority critical because rn in Skyra Lounge the announcement channel is #testing so we cannot make any announcements until this is merged


kyranet (changes requested):
Please split the serializer.type-channel.type checks into a separate function that returns a boolean.

kyranet (approved):

PR #1357, opened 2020-10-24
feat(routes): add route for public guild data

favna:
This will be required for the upcoming Dashboard rewrite.

This is retrieved server-sided so it can be used as meta tags, therefore there is no authentication available.
As far as I know the exposed fields are not of a sensitive nature so I think it should be fine.

### DIFF
--- a/src/lib/util/Models/ApiTransform.ts
+++ b/src/lib/util/Models/ApiTransform.ts
@@ -34,7 +34,6 @@ export function flattenGuild(guild: Guild): FlattenedGuild {
 		joinedTimestamp: guild.joinedTimestamp,
 		mfaLevel: guild.mfaLevel,
 		name: guild.name,
-		nameAcronym: guild.nameAcronym,
 		ownerID: guild.ownerID,
 		partnered: guild.partnered,
 		preferredLocale: guild.preferredLocale,
@@ -70,7 +69,6 @@ export interface FlattenedGuild
 		| 'joinedTimestamp'
 		| 'mfaLevel'
 		| 'name'
-		| 'nameAcronym'
 		| 'ownerID'
 		| 'partnered'
 		| 'preferredLocale'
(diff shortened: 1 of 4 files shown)


src/lib/util/Models/ApiTransform.ts
```diff
@@ -1,80 +1,94 @@
...
 		explicitContentFilter: guild.explicitContentFilter,
-		mfaLevel: guild.mfaLevel,
+		features: guild.features,
+		icon: guild.icon,
+		id: guild.id,
 		joinedTimestamp: guild.joinedTimestamp,
-		defaultMessageNotifications: guild.defaultMessageNotifications,
+		mfaLevel: guild.mfaLevel,
+		name: guild.name,
+		nameAcronym: guild.nameAcronym,
```
kyranet: This is not necessary, we calculate this in the dashboard.
```suggestion
```

src/lib/util/Models/ApiTransform.ts
```diff
@@ -1,80 +1,94 @@
...
+		| 'description'
+		| 'embedEnabled'
+		| 'explicitContentFilter'
+		| 'features'
+		| 'icon'
+		| 'id'
+		| 'joinedTimestamp'
+		| 'mfaLevel'
+		| 'name'
+		| 'nameAcronym'
```
kyranet:
```suggestion
```

src/routes/oauth/oauthUser.ts
```diff
@@ -101,15 +103,19 @@ export default class extends Route {
 							joinedTimestamp: null,
 							mfaLevel: 0,
 							name: oauthGuild.name,
+							nameAcronym: oauthGuild.name.slice(0, 2),
```
kyranet:
```suggestion
```

kyranet (dismissed):

kyranet (approved):

PR #1374, opened 2020-11-15
chore(dependencies): add missing deps to improve ws

vladfrangu:


favna:
This is looong loooong ago but I remember bufferutil actually decreasing performance in DJS at that time. Does it really improve performance these days?

vladfrangu:
Where's the claims in that / benchmarks..? Because like, it doesn't affect d.js directly, but ws itself

image

favna:
Gee I honestly don't remember / cannot provide evidence. Heck it might've even been in Discord chat. The only thing I can remember is that it was one of the core DJS devs that said to remove it at the time.

favna (approved):

PR #1385, opened 2020-11-19
feat: added guild fetch queue

### LINKED ISSUE #1285: request: fetch guild members after start-up and after joining a guild

One unexpected side effect from #1264 is that Skyra does not longer receive members at start-up, this makes her boot time so fast.

However, Skyra needs this data to report when users leave a guild, and having the cache of members is useful because it allows us to reduce the amount of member fetches we need at arbitrary times.

There are also existing commands that can benefit from this:

- `dehoist`

And future ones we may implement someday:

- `joinposition` (reads how many users joined before the user).
- `roleinfo members` (URD with the members who have a role).

One thing I was thinking of was an interval checking for a queue that does up to 30 fetches a time (pool implementation), and checks for guilds that are pending for fetching (`guild.members.cache.size !== guild.memberCount`), and tracks the last time it was synchronized (`WeakMap<Guild, number>` if we go for the extendable route).

We'll probably need either a manager (`guild.memberSynchronizer`?) or an extendable (`guild.members.sync()`) that does the fetch and sets the synchronization time.

Furthermore we'll need to implement our version of GuildMemberManager#fetch, because the built-in one builds a Collection and rejects in guilds with +30k members due to "timeout", both of which are undesirable.

I'm open to suggestions.

**Additional context**

Skyra's memberLeave logs show the time a member joined, it's been very frequently just `Unknown` lately.

kyranet:


favna (approved):

src/events/guilds/rawGuildCreateMemberFetch.ts
```diff
@@ -0,0 +1,12 @@
+import { GatewayDispatchEvents, GatewayGuildCreateDispatch } from 'discord-api-types/v6';
+import { Event, EventStore } from 'klasa';
+
+export default class extends Event {
+	public constructor(store: EventStore, file: string[], directory: string) {
+		super(store, file, directory, { name: GatewayDispatchEvents.GuildCreate, emitter: store.client.ws });
+	}
+
+	public run(data: GatewayGuildCreateDispatch['d'], id: number) {
```
vladfrangu: rename id to shardID for clarification

src/events/guilds/rawGuildDeleteMemberFetch.ts
```diff
@@ -0,0 +1,12 @@
+import { GatewayDispatchEvents, GatewayGuildDeleteDispatch } from 'discord-api-types/v6';
+import { Event, EventStore } from 'klasa';
+
+export default class extends Event {
+	public constructor(store: EventStore, file: string[], directory: string) {
+		super(store, file, directory, { name: GatewayDispatchEvents.GuildDelete, emitter: store.client.ws });
+	}
+
+	public run(data: GatewayGuildDeleteDispatch['d'], id: number) {
```
vladfrangu: ^

vladfrangu (dismissed):

src/events/guilds/rawGuildCreateMemberFetch.ts
```diff
@@ -0,0 +1,12 @@
...
+		this.client.guildMemberFetchQueue.add(id, data.id);
```
favna:
```suggestion
	public run(data: GatewayGuildCreateDispatch['d'], shardID: number) {
		this.client.guildMemberFetchQueue.add(shardID, data.id);
```

src/events/guilds/rawGuildDeleteMemberFetch.ts
```diff
@@ -0,0 +1,12 @@
...
+		this.client.guildMemberFetchQueue.remove(id, data.id);
```
favna:
```suggestion
	public run(data: GatewayGuildDeleteDispatch['d'], shardID: number) {
		this.client.guildMemberFetchQueue.remove(shardID, data.id);
```

favna (changes requested):

vladfrangu (approved):

PR #1390, opened 2020-11-20
fix(unban,emoji): throw strings instead of promises

kyranet:


favna (approved):
I have no idea why this was even going wrong because commandError event awaits promises

PR #1414, opened 2020-11-29
feat: make dj, moderator, and administrator, multi-roles

kyranet:


favna (approved):
Approving under the assumption that you tested the migration with data

PR #1429, opened 2020-12-13
fix(guildentity): fixed saving custom commands from dashboard

favna:


src/lib/database/entities/GuildEntity.ts
```diff
@@ -66,7 +66,7 @@ export class GuildEntity extends BaseEntity {
 	@Column('varchar', { name: 'disabledCommands', length: 32, array: true, default: () => 'ARRAY[]::VARCHAR[]' })
 	public disabledCommands: string[] = [];
 
-	@ConfigurableKey({ dashboardOnly: true, description: LanguageKeys.Settings.DashboardOnlyKey, type: 'customcommand' })
+	@ConfigurableKey({ dashboardOnly: true, dashboardArray: true, description: LanguageKeys.Settings.DashboardOnlyKey, type: 'customcommand' })
```
kyranet: All you had to do is specify `array: true`.

kyranet (changes requested):

src/lib/database/entities/GuildEntity.ts
Favna: Wrong. That breaks TypeORM.

kyranet (approved):

PR #1438, opened 2020-12-29
Update en-US to use "DJ" instead of "deejay" or "dee-jay"

Tylertron1998:
a [DJ](https://en.wikipedia.org/wiki/Disc_jockey) is a **D**isk **J**ockey - deejay/dee-jay is incorrect as it's an abbreviation.

This also brings consistency as 'DJ' was used in other places.


Tylertron1998:
Closed in favour of using https://translation.skyra.pw/

PR #1461, opened 2021-01-12
feat: add more precise i18next options

Quantumlyy:


package.json
```diff
@@ -91,9 +92,9 @@
 		"klasa-dashboard-hooks": "skyra-project/klasa-dashboard-hooks#master",
 		"node-fetch": "^2.6.1",
 		"pg": "^8.5.1",
-		"tslib": "1.11.2",
+		"tslib": "2.1.0",
```
kyranet: Revert

package.json
```diff
@@ -91,9 +92,9 @@
...
 		"tsyringe": "^4.4.0",
-		"typeorm": "0.2.28",
+		"typeorm": "0.2.29",
```
kyranet: Revert

kyranet (changes requested):

src/lib/util/constants.ts
```diff
@@ -297,13 +297,17 @@ export const clientOptions: Partial<KlasaClientOptions> = {
...
-		i18next: {
-			preload: [...handlers.keys()],
+		i18next: (_: string[], languages: string[]) => ({
+			supportedLngs: languages,
+			preload: languages,
 			returnObjects: true,
 			returnEmptyString: false,
 			returnNull: false,
 			load: 'all',
+			lng: 'en-US',
```
kyranet, line 307: Is this necessary?

Quantumlyy: Not necessary but at the same time, [the docs](https://www.i18next.com/overview/configuration-options) say it only affects automatic detection.

kyranet (approved):

PR #1471, opened 2021-01-13
fix: member joined timestamp not working

kyranet:
According to regex101, the new RegExp works: https://regex101.com/r/envtaw/2


kyranet:
That fixed it!
image

vladfrangu (approved):

Stitch07 (approved):

PR #1481, opened 2021-01-14
fix(gsearch): properly parse URL for images

favna:


vladfrangu (dismissed):

src/commands/Google/gsearch.ts
```diff
@@ -49,9 +49,11 @@ export default class extends RichDisplayCommand {
 			display.addPage((embed: MessageEmbed) => {
 				embed.setTitle(item.title).setURL(item.link).setDescription(item.snippet);
 
-				const imageUrl = item.pagemap?.cse_image?.find((image) => IMAGE_EXTENSION.test(image.src) && parseURL(image.src))?.src ?? '';
+				const imageUrl = parseURL(
+					item.pagemap?.cse_image?.find((image) => IMAGE_EXTENSION.test(image.src) && parseURL(image.src))?.src ?? ''
+				);
```
kyranet: This should be a method, or if you want to extend https://github.com/skyra-project/skyra/blob/main/src/lib/util/iterator.ts:

```typescript
const imageUrl = first(map((item.pagemap?.cse_image ?? []).values(), (image) => parseURL(image.src)), (url) => url !== null && IMAGE_EXTENSION.test(url.pathname));
```

*If only the pipeline operator was there...*

```typescript
const imageUrl = (item.pagemap?.cse_image ?? []).values()
	|> (_) => map(_, (image) => parseURL(image.src))
	|> (_) => first(_, (url) => url !== null && IMAGE_EXTENSION.test(url.pathname);
```

Nevertheless, the cleanest and most efficient way:

```typescript
// In a method
for (const image of item.pagemap?.cse_image ?? []) {
	const url = parseURL(image.src);
	if (url && IMAGE_EXTENSION.test(url.pathname)) return url.href;
}
```

kyranet (changes requested):

kyranet (changes requested):
Your unit tests also failed.

kyranet (approved):
Goole lol

PR #1507, opened 2021-01-20
readme: add translation section

Rexogamer:
It does what it says on the tin.

*Things to consider:*
- Should we look into any sort of Crowdin badge (like [this](https://github.com/badges/shields/issues/4906))?
- Should we add the Crowdin logo to the side like [this?](https://github.com/skyra-project/skyra-sharp/blob/master/README.md#translations-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,9 @@ Klasa.
 
 _Who knows, I might write a book someday for Skyra's lore._
 
+## Translating Skyra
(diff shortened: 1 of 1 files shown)


kyranet:
> - Should we look into any sort of Crowdin badge (like [this](https://github.com/badges/shields/issues/4906))?
> - Should we add the Crowdin logo to the side like [this?](https://github.com/skyra-project/skyra-sharp/blob/master/README.md#translations-)

Yes and yes.

kyranet (approved):

PR #1488, opened 2021-01-16
refactor: switch RichDisplay to PaginatedMessage

kyranet:
| ✔️ ~~This Pull Request needs https://github.com/sapphire-project/utilities/pull/60 to be approved, merged, and released beforehand. Once the linked Pull Request is merged, this one will receive its last commit to update `@sapphire/discord.js-utilities`.~~ |
| --- |

| ⚠️ This Pull Request needs https://github.com/sapphire-project/utilities/pull/62 to be approved, merged, and released beforehand. Once the linked Pull Request is merged, this one will receive its last commit to update `@sapphire/discord.js-utilities`. |
| --- |

PR #1509, opened 2021-01-21
chore: switch to PM

### LINKED ISSUE #1465: request: replace RichDisplay with PaginatedMessage

RichDisplay is a Klasa-specific thing, but not only that, it's also flawed in many ways.

[`@sapphire/discord.js-utilities`](https://www.npmjs.com/package/@sapphire/discord.js-utilities) has `PaginatedMessage`, which has all the features we need.

None.

**Additional context**

N/a.

kyranet:
Supersedes #1488
Fixes #1465


src/commands/Management/triggers.ts
```diff
@@ -119,17 +119,18 @@ export default class extends SkyraCommand {
...
+		});
 
 		for (const page of chunk(output, 10)) {
-			display.addPage((embed: MessageEmbed) => embed.setDescription(page));
+			display.addPageEmbed((embed) => embed.setDescription(page));
 		}
 
-		return display.start(message, undefined, { time: 120000 });
+		display.setIdle(120000);
+		return display.start(message);
```
vladfrangu: Use Time enum, or this
```suggestion
		return display.setIdle(120_000).start(message);
```

src/lib/structures/UserPaginatedMessage.ts
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	/**
+	 * Sets up the message's reactions and the collector.
+	 * @param channel The channel the handler is running at.
+	 * @param author The author the handler is for.
+	 */
+	protected async setUpReactions(channel: TextChannel | NewsChannel, author: User): Promise<void> {
+		if (this.pages.length > 1) await super.setUpReactions(channel, author);
+	}
```
vladfrangu, line 64: Can't this be merged in upstream?

src/lib/structures/UserPaginatedMessage.ts
```diff
@@ -0,0 +1,135 @@
...
+			video: embed.video ?? template.video ?? undefined,
+			footer: embed.footer ?? template.footer ?? undefined
+		};
+	}
+
+	private mergeArrays<T>(template?: T[], array?: T[]): undefined | T[] {
+		if (!array) return template;
+		if (!template) return array;
+		return [...template, ...array];
+	}
```
vladfrangu, line 118: probably best as a util

vladfrangu (changes requested):

src/lib/structures/UserPaginatedMessage.ts
kyranet: wdym?

src/lib/structures/UserPaginatedMessage.ts
kyranet: Probably in another PR, I cba to write tests for those utilities, specially with the amount of cases they handle.

src/lib/structures/UserPaginatedMessage.ts
vladfrangu: aka move the if check in the main module

kyranet: If that behaviour is desirable, I can move it upstream.

However... what if somebody defines custom actions which do page-independent actions? Say an info page.

vladfrangu: Well that'd be a different page so.. 🤔

kyranet: Well... nothing stops them from extending the behaviour to add virtual pages which aren't on the `this.pages` array.

vladfrangu: Riight, but... what's the point of reacting with one page? imo, if you want that behavior, you WANT to extend anyways

favna (approved):

PR #1533, opened 2021-01-27
fix(api): fix auth for real

favna:
| ⚠️ This Pull Request needs https://github.com/skyra-project/skyra.pw/pull/229 to be approved and merged beforehand. |
| --- |

### **_TODO:_**

- [ ] **_For some reason `auth` is `null` in `request.auth` in the `@authenticated` header and I have no frigging clue why_**

### DIFF
--- a/src/lib/api/utils.ts
+++ b/src/lib/api/utils.ts
@@ -1,16 +1,124 @@
 import { GuildSettings } from '#lib/database';
 import { hasAtLeastOneKeyInMap } from '#utils/comparators';
-import { Guild, GuildMember, Permissions } from 'discord.js';
+import { Store } from '@sapphire/pieces';
+import type { ApiRequest, ApiResponse, LoginData } from '@sapphire/plugin-api';
+import { createFunctionInhibitor } from '@skyra/decorators';
+import type { RESTAPIPartialCurrentUserGuild } from 'discord-api-types/v8';
+import { Client, Guild, GuildMember, Permissions } from 'discord.js';
+import { RateLimitManager } from 'klasa';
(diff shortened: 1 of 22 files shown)


src/lib/api/utils.ts
```diff
@@ -1,16 +1,124 @@
...
 	const [roles, pnodes] = await guild.readSettings((settings) => [settings[GuildSettings.Roles.Admin], settings.permissionNodes]);
 
 	return isAdmin(member, roles) && (pnodes.run(member, 'conf') ?? false);
 }
+
+export async function getManageable(id: string, oauthGuild: RESTAPIPartialCurrentUserGuild, guild: Guild | undefined): Promise<boolean> {
+	if (oauthGuild.owner) return true;
+	if (typeof guild === 'undefined') return new Permissions(Number(oauthGuild.permissions)).has(Permissions.FLAGS.MANAGE_GUILD);
+
+	const member = await guild.members.fetch(id).catch(() => null);
```
favna, line 65: Note that we really do need to fetch. In dev I was missing various guilds because they weren't cached yet.

src/lib/util/constants.ts
```diff
@@ -298,6 +300,11 @@ export const clientOptions: Partial<ClientOptions> = {
 	logger: {
 		instance: new Logger({ level: DEV ? LogLevel.Debug : LogLevel.Info })
 	},
+	api: {
+		auth: ({
+			transformers: [transformOauthGuildsAndUser]
```
favna, line 305: Because for some goddamn reason it creates a circular or something when in the config file causing everything to crash and burn

kyranet (approved):

PR #1567, opened 2021-02-17
feat: added Skyra.Database scaffold

kyranet:

PR #1575, opened 2021-02-19
chore: add base for all .NET projects

Tylertron1998:
This PR will add the base solution and required extras for all .NET services going forward.
It also closes #1567 (while that PR is still needed, I'll reopen a PR later to add that stuff back in, instead based off this).
This should have been done from the start as this gives us a proper base to start new services from, instead of being lumped in (and therefore requiring other service PRs to wait for) #1567.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,6 +46,5 @@ gh
 wiki/
 
 # Ignore dotnet build outputs
-
 obj/
 bin/
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,8 @@ gh
 
 # Ignore the "wiki" folder so we can checkout the wiki inside the same folder
 wiki/
+
+# Ignore dotnet build outputs
+
+obj/
+bin/
--- a/services/Directory.Build.props
+++ b/services/Directory.Build.props
@@ -7,11 +7,6 @@
 
     <!-- We add InternalsVisibleTo to every project, making sure Skyra.Tests can see internals (to test) APART from Skyra.Tests itself. -->
 
-    <ItemGroup>
-        <MyProjectReferences Include="**\*.*proj" />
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -43,4 +43,9 @@ package-lock.json
 gh
 
 # Ignore the "wiki" folder so we can checkout the wiki inside the same folder
-wiki/
\ No newline at end of file
+wiki/
+
+# Ignore dotnet build outputs
+
```
favna:
```suggestion
```

favna (changes requested):

favna (approved):

services/Directory.Build.props
```diff
@@ -0,0 +1,15 @@
+<Project>
+    <PropertyGroup>
+        <Deterministic>true</Deterministic>
+        <Features>strict</Features>
```
Tylertron1998, line 4: note: we use deterministic builds here because that means the compiler will produce the exact same binary output (and be sped up by a factor of 70%) which is good in situations like testing, where the actual project doesn't change, but the tests do (or really any consumer of a library) - and strict features to remove some intentional bugs in the compiler (microsoft added intentional bugs, replicating old bugs from the old compiler for backwards compat, cheers MS!)

services/Skyra.Tests/UnitTest1.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+	public class Tests
+	{
+		[SetUp]
+		public void Setup()
+		{
+		}
+
+		[Test]
+		public void Test1()
```
Tylertron1998, line 13: we'll eventually write actual tests (@Favna can help me scream at kyra to add tests as we go)

kyranet (approved):

PR #1582, opened 2021-02-20
feat: added Skyra.Database and Skyra.Grpc

kyranet:


assets/protos/greet.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto3";
+
+option csharp_namespace = "Skyra.Grpc.Services";
+
+package greet;
+
+// The greeting service definition.
+service Greeter {
```
Tylertron1998: do we need the greeter service? it's an easy way to test gRPC connections, but on the other hand, some of the other endpoints (such as adding points) are really easy to test too (via bloomrpc or some other testing client)

assets/protos/member.proto
```diff
@@ -0,0 +1,29 @@
...
+
+// The greeting service definition.
+service Member {
+  // Sends a greeting
+  rpc AddPoints (Points) returns (Result);
+  rpc GetPoints (MemberQuery) returns (Result);
+}
+
+// The request message containing the user's name.
+message Points {
```
Tylertron1998: I think we should have some naming conventions for gRPC types - I am personally in favor of [Type]{Request | Result} - in this case, it'd be `PointsRequest` and the endpoint would return a `PointsResult`

services/Skyra.Database/Extensions/DbSetExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+	{
+		public async static Task<T> UpsertAsync<T>(this DbSet<T> db, object key, Func<T> data) where T : class
+		{
+			var entity = await db.FindAsync(key);
+
+			if (entity is null)
+			{
+				entity = data();
+				await db.AddAsync(entity);
+			}
```
Tylertron1998, line 17:
```suggestion
			entity ??= data();
			await db.AddAsync(entity);
```

services/Skyra.Database/Models/Banner.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.ComponentModel.DataAnnotations;
+using System.ComponentModel.DataAnnotations.Schema;
+using Microsoft.EntityFrameworkCore;
+
+#nullable disable
+
+namespace Skyra.Database.Models
+{
+	[Table("banner")]
+	public partial class Banner
```
Tylertron1998: any reason why these are partial classes?

services/Skyra.Grpc/Services/GreeterService.cs
```diff
@@ -0,0 +1,24 @@
+using System.Threading.Tasks;
+using Grpc.Core;
+using Microsoft.Extensions.Logging;
+
+namespace Skyra.Grpc.Services
+{
+	public class GreeterService : Greeter.GreeterBase
```
Tylertron1998: same as above - do we still need the greeter service?

services/Skyra.Grpc/Startup.cs
```diff
@@ -0,0 +1,49 @@
...
+			if (env.IsDevelopment())
+			{
+				app.UseDeveloperExceptionPage();
+			}
+
+			app.UseRouting();
+
+			app.UseEndpoints(endpoints =>
+			{
+				endpoints.MapGrpcService<GreeterService>();
```
Tylertron1998: same here, do we need it?

Tylertron1998 (commented):
other then these few things which may need addressing, looks good to me

services/Skyra.Grpc/Startup.cs
```diff
@@ -0,0 +1,49 @@
...
+namespace Skyra.Grpc
+{
+	public class Startup
+	{
+		// This method gets called by the runtime. Use this method to add services to the container.
+		// For more information on how to configure your application, visit https://go.microsoft.com/fwlink/?LinkID=398940
+		public void ConfigureServices(IServiceCollection services)
+		{
+			services.AddGrpc();
+			services.AddSingleton<SkyraDbContext>();
```
Tylertron1998: This should be scoped - not singleton, each request should have a new dbcontext otherwise there will be out of date issues & race conditions

```suggestion
			services.AddScoped<SkyraDbContext>();
```

Tylertron1998 (changes requested):

services/Skyra.Database/Models/Banner.cs
kyranet: Ask EFCore, not me.

Tylertron1998: does rider find any other partial classes for any of these? if not, it'll be a good idea to make them sealed, not partial

assets/protos/member.proto
kyranet: We're using `Query` just below, should I rename it to `Request`?

services/Skyra.Database/Extensions/DbSetExtensions.cs
kyranet: I don't think it's wise to always force an `AddAsync`.

Tylertron1998: Good point!

assets/protos/member.proto
Tylertron1998: imo, we're going to be adding a load of other services later down the line - and if we have each of them have a message named `Query` and also a message named `Request` - we're going to have about 20 different classes named `Query` and `Request` - which is going to reallllly bloat intellisense, so prefixing them with a sensible name is better.

services/Skyra.Database/Models/Banner.cs
kyranet: That wouldn't be wise, in fact, there are keys that are `virtual` (EFCore's relationships).

I ran Rider's code cleanup on the whole solution, it removed the partials.

Tylertron1998: perfect!

assets/protos/member.proto
```diff
@@ -0,0 +1,29 @@
...
+
+option csharp_namespace = "Skyra.Grpc.Services";
+
+package member;
+
+// The greeting service definition.
+service Member {
+  // Sends a greeting
+  rpc AddPoints (PointsQuery) returns (Result);
+  rpc GetPoints (MemberQuery) returns (Result);
```
Tylertron1998:
```suggestion
  rpc AddPoints (PointsQuery) returns (PointsResult);
  rpc GetPoints (MemberQuery) returns (PointsResult);
```

assets/protos/member.proto
```diff
@@ -0,0 +1,29 @@
...
+message MemberQuery {
+  string id = 1;
+}
+
+// The response message containing the greetings.
+message Result {
+  bool success = 1;
+  int64 amount = 2;
+  string error_message = 3;
+}
```
Tylertron1998:
```suggestion
message PointsResult {
  bool success = 1;
  int64 amount = 2;
  string error_message = 3;
}
```

services/Skyra.Grpc/Services/MemberService.cs
```diff
@@ -0,0 +1,66 @@
...
+			try
+			{
+				var user = await _context.Users.FindAsync(request.Id);
+
+				if (user is null)
+					return new Result
+					{
+						Success = true,
+						Amount = 0
+					};
```
Tylertron1998:
```suggestion
				if (user is null)
				{
					return new Result
					{
						Success = true,
						Amount = 0
					};
				}
```

Tylertron1998 (changes requested):
a few more changes, then LGTM (for real this time)

services/Skyra.Grpc/Services/MemberService.cs
kyranet: Actually, I have a better idea.

Tylertron1998 (dismissed):

Tylertron1998 (dismissed):

Tylertron1998 (approved):

favna (approved):

PR #1612, opened 2021-03-03
feat: added gRPC into Node.js

kyranet:
---

closes #1763
closes #1761


Tylertron1998 (changes requested):
please write tests, I didn't spend hours writing tests for the services for this to not be tested also

kyranet:
> please write tests, I didn't spend hours writing tests for the services for this to not be tested also

There, cross-language e2e implemented in 5dd11ec.

tests/lib/Grpc.test.ts
```diff
@@ -0,0 +1,30 @@
+import { Store } from '@sapphire/pieces';
+
+const { grpc } = Store.injectedContext;
+
+describe('Grpc', () => {
```
Tylertron1998: I feel like it's probably a smart idea to wipe the db before every run again, as we do in Skyra.Grpc tests - otherwise tests depend on a fresh database, which might not be the case for some development environments

tests/lib/Grpc.test.ts
```diff
@@ -0,0 +1,30 @@
...
+	test('waitForReady succeeds', async () => {
+		await expect(grpc.waitForReady()).resolves.toBe(undefined);
+	});
+
+	test('Member#addPoints', async () => {
```
Tylertron1998: GIVEN/WHEN/THEN please

Tylertron1998 (changes requested):

tests/lib/Grpc.test.ts
kyranet: Not possible, you didn't add something to nuke the entire DB, and all DB stuff will be removed from the Node.js code as soon as Skyra.Grpc covers all the necessary methods.

Not that we should have such a dangerous command, tho.

tests/lib/Grpc.test.ts
```diff
@@ -0,0 +1,30 @@
...
+		const result = await grpc.members.addPoints({ id: '1', amount: 200 });
+		expect(result.success).toBe(true);
+		expect(result.amount).toBe(250);
+		expect(result.errorMessage).toBe('');
+	});
+
+	test('Member#getPoints', async () => {
+		const result = await grpc.members.getPoints({ id: '1' });
+		expect(result.success).toBe(true);
+		expect(result.amount).toBe(250);
```
Tylertron1998: you're relying on the earlier tests' result (adding 250 points to the member with id of '1') here, which is not good at all - you can't rely on other tests results for these tests.

Tylertron1998 (changes requested):

tests/lib/Grpc.test.ts
kyranet: You didn't add setPoints nor resetPoints.

scripts/grpc.mjs
```diff
@@ -0,0 +1,105 @@
...
+	console.log(gray(`└─ ${directory}${sep}${green(last)}`));
+}
+
+async function prepareGeneratedDirectory() {
+	await mkdir(generatedDirectory, { recursive: true });
+}
+
+function isWindows() {
+	return process.platform === 'win32';
+}
```
vladfrangu: No reason for this to be a function

vladfrangu (changes requested):

tests/lib/Grpc.test.ts
```diff
@@ -0,0 +1,90 @@
...
+				await grpc.members.addPoints({ guildId: '1', userId: '1', amount: 100 });
+				const result = await grpc.members.getPoints({ guildId: '1', userId: '1' });
+				expect(result.status).toBe(Status.SUCCESS);
+				expect(result.experience).toBe(100);
+			});
+		});
+
+		describe('Reset Points', () => {
+			test('GIVEN a non-existing entry THEN returns 0', async () => {
+				const resetResult = await grpc.members.resetPoints({ guildId: '1', userId: '1' });
```
Tylertron1998: is it possible to use constants here for the id's (like we do in .NET tests?)

Tylertron1998 (changes requested):
please add a test for a call that returns `Status.FAILED` (you can achieve this by doing a member query where the guild id is longer then 19 chars)

Tylertron1998 (dismissed):
🥚cellent

vladfrangu (dismissed):
Looks..good to merge I SUPPOSE

src/.env.development
```diff
@@ -14,6 +14,8 @@ API_ORIGIN='http://localhost.org:3000'
 API_PREFIX='/'
 API_PORT=1234
 
+GRPC_ADDRESS='localhost:5000'
```
favna:
```suggestion
GRPC_ADDRESS='localhost:8291'
```

Should match upcoming PR for running Grpc in Docker (which will be development default, overwrite it in `.env.development.local` if you run it from command line)

src/.env
```diff
@@ -14,6 +14,8 @@ API_ORIGIN='http://localhost.org:3000'
 API_PREFIX='/'
 API_PORT=1234
 
+GRPC_ADDRESS='localhost:5000'
```
favna:
```suggestion
GRPC_ADDRESS='localhost:8291'
```

favna (changes requested):

.github/workflows/continuous-integration.yml
```diff
@@ -79,6 +79,19 @@ jobs:
...
+        env:
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: skyra-nodejs-test
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 5432:5432
```
favna: Needs to have Grpc added, so this PR depends on the `feat/docker-dotnet` branch.

.github/workflows/continuous-integration.yml
```diff
@@ -97,6 +110,19 @@ jobs:
...
+        run: |
+          dotnet new tool-manifest
+          dotnet tool install dotnet-ef
+          dotnet tool run dotnet-ef database update --project services/Skyra.Database
+          dotnet run --project services/Skyra.Grpc &
+          while ! nc -z localhost 5000; do
+            sleep 0.1 # wait for 1/10 of the second before check again
+          done
+        env:
+          POSTGRES_DB: skyra-nodejs-test
```
favna: To be removed when Grpc service is added

package.json
```diff
@@ -30,8 +31,9 @@
 		"test": "jest",
 		"test:watch": "jest --watch",
 		"update": "yarn upgrade-interactive --latest",
-		"build": "tsc -b src",
+		"build": "yarn scripts:grpc && tsc -b src",
```
favna:
```suggestion
		"prebuild": "yarn scripts:grpc",
		"build": "tsc -b src",
```

package.json
```diff
@@ -114,21 +118,24 @@
...
 		"@types/jest": "^26.0.22",
 		"@types/node": "^14.14.37",
-		"@types/node-fetch": "^2.5.9",
+		"@types/node-fetch": "^2.5.10",
 		"@types/pg": "^7.14.11",
 		"@types/ws": "^7.4.1",
-		"@typescript-eslint/eslint-plugin": "^4.20.0",
-		"@typescript-eslint/parser": "^4.20.0",
+		"@typescript-eslint/eslint-plugin": "^4.21.0",
+		"@typescript-eslint/parser": "^4.21.0",
```
favna: Revert lest this is going to cause conflicts down the road, i.e. if not merged before sunday and depfu comes around.

package.json
```diff
@@ -139,7 +146,7 @@
 		"pretty-quick": "^3.1.0",
 		"ts-jest": "^26.5.4",
 		"ts-node": "^9.1.1",
-		"typescript": "4.2.3"
+		"typescript": "4.2.4"
```
favna: Revert lest this is going to cause conflicts down the road, i.e. if not merged before sunday and depfu comes around.

package.json
```diff
@@ -114,21 +118,24 @@
...
 		"@types/diff": "^5.0.0",
+		"@types/google-protobuf": "^3.7.4",
 		"@types/he": "^1.1.1",
 		"@types/i18next-fs-backend": "^1.0.0",
-		"@types/ioredis": "^4.22.2",
+		"@types/ioredis": "^4.22.3",
 		"@types/jest": "^26.0.22",
 		"@types/node": "^14.14.37",
-		"@types/node-fetch": "^2.5.9",
+		"@types/node-fetch": "^2.5.10",
```
favna: Revert lest this is going to cause conflicts down the road, i.e. if not merged before sunday and depfu comes around.

package.json
```diff
@@ -114,21 +118,24 @@
 		"@skyra/saelem": "^3.0.2",
 		"@types/backoff": "^2.5.1",
 		"@types/diff": "^5.0.0",
+		"@types/google-protobuf": "^3.7.4",
 		"@types/he": "^1.1.1",
 		"@types/i18next-fs-backend": "^1.0.0",
-		"@types/ioredis": "^4.22.2",
+		"@types/ioredis": "^4.22.3",
```
favna, line 80: Revert lest this is going to cause conflicts down the road, i.e. if not merged before sunday and depfu comes around.

package.json
```diff
@@ -92,8 +95,9 @@
 		"discord.js": "^12.5.3",
 		"dotenv-cra": "^2.0.2",
 		"gifencoder": "^2.0.1",
+		"google-protobuf": "^3.16.0-rc.1",
 		"he": "^1.2.0",
-		"ioredis": "^4.25.0",
+		"ioredis": "^4.26.0",
```
favna: Revert lest this is going to cause conflicts down the road, i.e. if not merged before sunday and depfu comes around.

scripts/grpc.mjs
```diff
@@ -0,0 +1,106 @@
+import { execFile } from 'child_process';
+import { gray, green, red } from 'colorette';
+import { mkdir, readdir, rm, writeFile } from 'fs/promises';
+import { dirname, join, sep } from 'path';
+import { fileURLToPath } from 'url';
+import { promisify } from 'util';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
```
favna, line 9: Remove and replace with `new URL` stuff to get all the directories and write all the files.

src/.env.test
```diff
@@ -1 +1,3 @@
+PGSQL_DATABASE_NAME='skyra-nodejs-test'
```
favna: And what happens, like in 99% of the cases when running tests locally, where this DB doesn't exist and the tests are run? If tests break in such a case like I suspect then find a better solution for this.

In fact, "Node -> Grpc" is an integration test, but you're hijacking the unit tests package with the tests now. "Node -> Grpc" tests should be its own complete package with its own `jest.config`, and its own `package.json`, the whole shebang.

tests/lib/Grpc.test.ts
```diff
@@ -0,0 +1,101 @@
+import { ResponseError, Status } from '#lib/grpc';
```
favna: See other comment as to why this file should not be here

favna (changes requested):

scripts/grpc.mjs
kyranet: This script depends on `execFile`, which does not support `URL`s:

image

favna (dismissed):

favna (dismissed):

PR #1630, opened 2021-03-07
fix(dice): default to 1 instead of sides

Stitch07:


src/commands/Fun/dice.ts
```diff
@@ -41,15 +41,15 @@ export class UserCommand extends SkyraCommand {
 		let modifier = 0;
 		if (typeof pattern === 'number') {
 			if (!isNumber(pattern) || pattern <= 0) this.error(LanguageKeys.Serializers.InvalidInt, { name: 'dice' });
-			amount = pattern;
+			amount = 1;
 			dice = 6;
```
kyranet: At least, use `pattern` if you're going to translate a number for a dice's faces instead of the amount of rolls.
```suggestion
			dice = pattern;
```

kyranet (changes requested):

kyranet (approved):

favna (approved):

PR #1636, opened 2021-03-13
fix(announcement): resolved bugs with parsing ids

kyranet:


src/lib/util/util.ts
```diff
@@ -502,6 +502,41 @@ export function cleanMentions(guild: Guild, input: string) {
  */
 export const extractMentions = (input: string) => input.match(/<(@[!&]?)(\d{17,19})>/g) ?? [];
```
favna, line 503: I don't think we use this anywhere else so if that's the case remove it.

favna (approved):

favna (changes requested):
1 more bug to fix (See discord). Just doing this so it cannot accidentally be merged.

favna (approved):

PR #1658, opened 2021-03-19
feat(custom-commands): add empty formatter

kyranet:
This is useful for defining arguments at the start of a tag (to specify their order) but not show them, for example:
`{target:user | empty}{money:integer | empty}{money:integer} dollar(s) was sent from {author} to {target:user}`


vladfrangu (approved):
k

favna (approved):

PR #1660, opened 2021-03-19
fix(birthday): handle unavailable guild and correct permissions

kyranet:


src/tasks/removeBirthdayRole.ts
```diff
@@ -1,16 +1,39 @@
...
 export class UserTask extends Task {
+	private readonly requiredPermissions = new Permissions(Permissions.FLAGS.MANAGE_ROLES);
+
 	public async run(data: RemoveBirthdayRoleData): Promise<PartialResponseValue | null> {
+		// Get and check the guild:
 		const guild = this.context.client.guilds.cache.get(data.guildID);
 		if (!guild) return null;
 
+		// If the guild is not available, re-schedule the task by creating
+		// another with the same data but happening 20 seconds later.
```
favna:
```suggestion
		// another with the same data but happening 30 seconds later.
```

favna (changes requested):

PR #1665, opened 2021-03-19
refactor(db): rename all keys to `lower-kebab-case`

kyranet:
- Added case-insensitivity support for the `conf` command.
- Added auto-aliases to allow skipping dashes (`-`) in configuration names.
- Refactored all keys from the `public.guilds` table to be `lower-kebab-case` instead of mixing it with `camelCase`.
- Updated .NET code to match new keys.


src/lib/database/migrations/1616168146993-V39_RenamePropertiesToSnakeCase.ts
```diff
@@ -0,0 +1,99 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class V39RenamePropertiesToSnakeCase1616168146993 implements MigrationInterface {
```
favna:
```suggestion
export class V39RenamePropertiesToKebabCase1616168146993 implements MigrationInterface {
```

Also update file name and reference in `SetMigrations` accordingly

favna (changes requested):
I haven't really checked if every column is covered but I assume you have.

favna (approved):

PR #1672, opened 2021-03-20
feat(giveaway): add `allowed_roles`

### LINKED ISSUE #1415: request: add allowedRoles filters into giveaways

Sometimes administrators want to make giveaways public for everyone to see, but locked to users with a role to participate.

Add an extra step in the `gcreate` command and an extra argument in the `gstart` command.

N/a.

**Additional context**

N/a.

kyranet:
Fixes #1415


src/lib/database/migrations/1616201639503-V42_AddAllowedRolesIntoGiveaways.ts
```diff
@@ -0,0 +1,20 @@
+import { MigrationInterface, QueryRunner, TableColumn } from 'typeorm';
+
+export class V42AddAllowedRolesIntoGiveaways1616201639503 implements MigrationInterface {
+	public async up(queryRunner: QueryRunner): Promise<void> {
+		await queryRunner.addColumn(
+			'giveaway',
+			new TableColumn({
+				name: 'allowed_roles',
```
favna, line 8: why is this snake cased and not kebab cased?

src/lib/database/entities/GiveawayEntity.ts
```diff
@@ -69,6 +73,9 @@ export class GiveawayEntity extends BaseEntity {
 	@Column('integer', { default: 1 })
 	public minimumWinners = 1;
 
+	@Column('varchar', { length: 19, array: true, default: () => 'ARRAY[]::VARCHAR[]' })
```
favna, line 76: why not kebab case?

favna (changes requested):

src/lib/database/migrations/1616201639503-V42_AddAllowedRolesIntoGiveaways.ts
kyranet: PostgreSQL's naming convention is to use `snake_case`, and all tables except `guilds` follow this.

The reason why I didn't for the guilds table, is because users find `kebab-case` easier, either way, it doesn't follow conventions per-se, as dots are precisely pretty against the PostgreSQL's standards (and more so that it has specific a meaning).

*It definitely didn't occur to me 2 hours later that I could have replaced `_` with `-` in the maps when generating the settings trees.*

favna (approved):

PR #1675, opened 2021-03-20
feat(birthday): add i18n-aware date parser

kyranet:


src/commands/Birthday/set-birthday.ts
```diff
@@ -45,38 +45,35 @@ export class UserCommand extends SkyraCommand {
...
+			return Args.error({
+				argument,
+				parameter,
+				identifier: LanguageKeys.Commands.Misc.SetBirthdayInvalidDate,
+				context: { formatWithYear: format, formatWithoutYear: removeYear(format) }
+			});
+		}
 
-		const year = result[1] === undefined ? null : Number(result[1]);
+		const year = result.groups!.year === undefined ? null : Number(result.groups!.year);
```
favna, line 66:
```suggestion
		const year = result.groups?.year === undefined ? null : Number(result.groups.year);
```

Not sure if that second `!` can be removed when making the first a `?` but I do think so

favna (changes requested):

src/commands/Birthday/set-birthday.ts
kyranet: `groups` being `Record<string, string> | undefined` is being so because it's generic for all RegExps, however, the RegExps we're using all use groups, and therefore, this can never be undefined.

favna (approved):

PR #1681, opened 2021-03-21
refactor(config): switch to .env

kyranet:


src/.env.development
```diff
@@ -0,0 +1,90 @@
+CLIENT_NAME="Skyra"
+CLIENT_VERSION="6.0.0-dev [Sapphire Edition]"
+CLIENT_PREFIX="sd!"
+CLIENT_REGEX_PREFIX=""
+CLIENT_OWNERS=""
+CLIENT_ID=""
+CLIENT_SHARDS="auto"
```
vladfrangu: There's a property d.js uses automatically, you should use that instead

src/lib/database/migrations/1594413973851-V01_MigrateClientStorage.ts
```diff
@@ -17,7 +16,14 @@ export class V01MigrateClientStorage1594413973851 implements MigrationInterface
 			new Table({
 				name: 'client',
 				columns: [
-					new TableColumn({ name: 'id', type: 'varchar', length: '19', default: `'${CLIENT_ID}'`, isNullable: false, isPrimary: true }),
+					new TableColumn({
```
vladfrangu, line 19: How do migrations get this?

vladfrangu (changes requested):

src/.env.development
kyranet: ... which property?

src/lib/database/migrations/1594413973851-V01_MigrateClientStorage.ts
kyranet: I added an `import '#root/config';` to the `database.config.ts` so envs are loaded before migrations.

src/events/ready.ts
```diff
@@ -1,12 +1,13 @@
...
 import { Slotmachine } from '#lib/games/Slotmachine';
 import { WheelOfFortune } from '#lib/games/WheelOfFortune';
 import { Events, Schedules } from '#lib/types/Enums';
-import { DEV, ENABLE_AUDIO, ENABLE_INFLUX, VERSION } from '#root/config';
 import { ApplyOptions } from '@sapphire/decorators';
 import { Event, EventOptions, Store } from '@sapphire/framework';
 import { blue, gray, green, magenta, magentaBright, red, white, yellow } from 'colorette';
 
-const style = DEV ? yellow : blue;
+const dev = process.env.NODE_ENV === 'development';
```
vladfrangu: already have a prop for this...and now they don't even check the same thing

vladfrangu (changes requested):

src/events/ready.ts
kyranet: What do you mean?

vladfrangu: `Client#dev` checks that the env isn't `production`. Consistency plz

kyranet: Because NODE_ENV has 3 possible values, `production`, `development`, and `test`.

Inside pieces, they're exclusively either `production` or `development`, since Jest doesn't load them.

kyranet: Either way, addressed that in 75a226a.

PR #1684, opened 2021-03-22
feat: add date support to timespan

kyranet:
- Adds date support in timespan.
  - Understandable as "time span until".
  - Allows things such as:
    - `s!remindme 2022-01-01 Wish my family a happy New Years`.
    - `s!ban @user 2024-01-01`.
- Removes YYYY and YYYY-MM overloads from usage, recommend YYYY-MM-DD for less ambiguity.
  - Also because in timespan, 2020 would be parsed as 2020 seconds, not as year 2020.


vladfrangu (approved):
If you tested, LGTM

PR #1688, opened 2021-03-22
feat(reputation): add reminder flag

kyranet:


vladfrangu (approved):
Making it a flag is meh (can't wait for slash commands). lGTM

PR #1690, opened 2021-03-24
feat(services): add Skyra.Cdn

kyranet:


services/Skyra.Cdn/Controllers/CdnController.cs
```diff
@@ -0,0 +1,77 @@
...
+	{
+		private readonly IDatabase _database;
+
+		public CdnController(IDatabase database)
+		{
+			_database = database;
+		}
+
+		[HttpGet("{id}")]
+		[ResponseCache(Duration = 60 * 60, Location = ResponseCacheLocation.Client, NoStore = false)]
```
Tylertron1998: Pls no magic numbers, make these constants (what does each 60 represent?)

services/Skyra.Cdn/Controllers/CdnController.cs
```diff
@@ -0,0 +1,77 @@
...
+		public async Task<IActionResult> Get(string id)
+		{
+			var result = await _database.GetAssetAsync(id);
+			if (!result.Success)
+			{
+				return NotFound();
+			}
+
+			var asset = result.Value!;
```
Tylertron1998, line 38: Use an actual null check here `if(x is not null)`

services/Skyra.Cdn/Controllers/CdnController.cs
```diff
@@ -0,0 +1,77 @@
...
+			{
+				return NotModified();
+			}
+
+			headers.LastModified = new DateTimeOffset(asset.LastModifiedAt);
+			headers.Set("ETag", asset.ETag);
+			return File(asset.Data, asset.ContentType);
+		}
+
+		private bool WasModified(Asset asset)
```
Tylertron1998: Could this be a local method?

services/Skyra.Cdn/Startup.cs
```diff
@@ -0,0 +1,44 @@
...
+
+		public IConfiguration Configuration { get; }
+
+		// This method gets called by the runtime. Use this method to add services to the container.
+		public void ConfigureServices(IServiceCollection services)
+		{
+			services.AddControllers();
+			services.AddScoped<SkyraDbContext>();
+			services.AddScoped<IDatabase, SkyraDatabase>();
+			services.AddSwaggerGen(c => { c.SwaggerDoc("v1", new OpenApiInfo {Title = "Skyra.Cdn", Version = "v1"}); });
```
Tylertron1998: Are the two "v1"'s here the ""same"" thing? If so, constant pls

Tylertron1998 (changes requested):

services/Skyra.Cdn/Controllers/CdnController.cs
kyranet: What?

services/Skyra.Cdn/Startup.cs
kyranet: That's... from the ASP.NET template, I have no idea what it is for

services/Skyra.Cdn/Controllers/CdnController.cs
kyranet: 60 seconds (1m) * 60, aka 60m, aka 1h, I wanted to make this more readable, but I'm unsure how, `Duration` takes an integer

services/Skyra.Cdn/Controllers/CdnController.cs
kyranet: `Response<T>.Value` is `T?`, being `null` when `Success` is false. Here, `Success` is true, and therefore, it cannot be null.

Tylertron1998: Hmmm... I wonder if we should change the design of this. Alright

services/Skyra.Cdn/Startup.cs
Tylertron1998: Please see if it can

services/Skyra.Cdn/Controllers/CdnController.cs
Tylertron1998: Two private const int's will do

kyranet: I can't use `TimeSpan.FromHours(1).Seconds` because it needs to be constant.

services/Skyra.Cdn/Startup.cs
kyranet: I just nuked it

services/Skyra.Cdn/Controllers/CdnController.cs
Tylertron1998: again - two constant integers `private const int _seconds = 60; private const int _minutes = 60;`

services/Skyra.Cdn/Controllers/CdnController.cs
Tylertron1998: Please make this a local method

PR #1691, opened 2021-03-25
refactor(moderation): rewrite for new features and enhancements

### LINKED ISSUE #1335: rewrite: moderation module improvements

kyranet:
Fixes #1335


kyranet:
Superseded by #2595

PR #1707, opened 2021-03-27
feat(error): add Sentry error logging

### LINKED ISSUE #1218: request: move all error logging to Sentry

A month ago, we have added Sentry, however, we haven't fully integrated it. As a result, we are still managing errors on Discord, which is bothersome as we sometimes have to delete a lot of messages from #production-errors.

However, Sentry gives us more control over this, including grouping errors, and this feels like a good solution.

- [ ] Switch the current error logging system with one based on Sentry.
- [ ] Add a webhook system so we are notified on Discord when errors happen.
- [ ] Send the error identifier so users can give us the steps they made to cause a command error.

N/a.

**Additional context**

N/a.

kyranet:
Fixes #1218. [Reference](https://docs.sentry.io/platforms/node/usage/#capturing-errors) for error capturing with Sentry.

- fix(analytics): properly unload events if INFLUX_ENABLED is false.
- fix(weeb): properly unload commands if WEEB_SH_TOKEN is unset.
- refactor(errors): rename `ApiError` to `Error`, the latter wasn't emitted and it only accepted strings.


src/lib/types/Augments.d.ts
```diff
@@ -94,6 +94,10 @@ declare module 'discord.js' {
 		emit(event: string | symbol, ...args: any[]): boolean;
 	}
 
+	interface ClientEvents {
+		[Events.TaskError]: [error: Error, payload: TaskErrorPayload];
+	}
```
favna, line 99: Can't this go in the block of custom emits above?

favna (changes requested):

src/lib/types/Augments.d.ts
kyranet: Not without sacrificing type checking in the Event (see that they're generic).

favna (approved):

PR #1715, opened 2021-03-30
fix(birthday): correct date format in extended help

kyranet:


src/languages/en-US/commands/misc.json
```diff
@@ -263,7 +263,7 @@
 		"explainedUsage": [
 			[
 				"BirthDate",
-				"Your birth-date, in the format YYYY/MM/DD. (Year/Month/Day)"
+				"Your birth-date, in the format $t(globals:dateFormat). ($t(globals:dateFormatExplanation))"
 			]
 		],
 		"examples": [
```
favna, line 269: examples is also hardcoded to YYYY/MM/DD

favna (changes requested):

favna (approved):

PR #1719, opened 2021-03-31
feat(leaderboard): make the message paginated

kyranet:
- feat(leader-board): make the message paginated
- refactor(leader-board): show external users as unknown
- refactor(routes): removed global leader board
- refactor(routes): show external users as unknown in local leader board

The reason why users are now shown as unknown when they're external, is to protect data leaks (such as users who use Skyra privately being shown in the global one), which can be very undesirable.

Other bots also show external members as unknown, too.


vladfrangu (dismissed):
Test it then LGTM

favna (dismissed):
UserLazyPaginatedMessage and UserPaginatedMessage should really be backported to Sapphire...

Also what Vladdy said.

kyranet:
> Test it then LGTM

It's tested.

---

Also, on a side note, since we're making this protect privacy, should we also remove global leaderboard and expose only the position?

cc: @Favna @vladfrangu

favna:
> Also, on a side note, since we're making this protect privacy, 

image

kyranet:
> > Also, on a side note, since we're making this protect privacy,
> 
> image

Please read the PR's description

favna:
> Also, on a side note, since we're making this protect privacy, should we also remove global leaderboard and expose only the position?

yes

favna (approved):

PR #1725, opened 2021-04-01
fix(suggest): reset emojis if they are unknown

kyranet:


favna (changes requested):
"reset emojis if they become unknown" but not using https://discord.js.org/#/docs/main/stable/class/Client?scrollTo=e-emojiDelete

wat

kyranet:
> "reset emojis if they become unknown" but not using [discord.js.org/#/docs/main/stable/class/Client?scrollTo=e-emojiDelete](https://discord.js.org/#/docs/main/stable/class/Client?scrollTo=e-emojiDelete)
> 
> wat

Out of the PR's scope, and emojis could be deleted anytime, even when Skyra is offline during maintenance/restart.

There's also the case in which users use emojis from external servers Skyra is also in.

favna (approved):

PR #1730, opened 2021-04-02
feat(logs): add new logs channels

kyranet:
- feat(settings): added `channels.logs.role-create`.
- feat(settings): added `channels.logs.role-update`.
- feat(settings): added `channels.logs.role-delete`.
- feat(settings): added `channels.logs.channel-create`.
- feat(settings): added `channels.logs.channel-update`.
- feat(settings): added `channels.logs.channel-delete`.
- feat(settings): added `channels.logs.emoji-create`.
- feat(settings): added `channels.logs.emoji-update`.
- feat(settings): added `channels.logs.emoji-delete`.
- feat(settings): added `channels.logs.server-update`.


vladfrangu (dismissed):
LGTM... but I will forever shame you for still keeping kebab-case column names when you should be using snake_case

vladfrangu (dismissed):
LGTM... but I will forever shame you for still keeping kebab-case column names when you should be using snake_case

~~Also my reviews don't count anymore, thx kyra~~

src/languages/en-US/guilds.json
```diff
@@ -0,0 +1,10 @@
+{
+	"explicitContentFilterDisabled": "Disabled",
+	"explicitContentFilterMembersWithoutRoles": "Members without roles",
+	"explicitContentFilterAllMembers": "All members",
+	"messageNotificationsAll": "All Messages",
+	"messageNotificationsMentions": "Only @mentions",
```
favna, line 6: Does this only get used in embeds or in raw text as well? Because in case of the latter if some server has a role called `mentions` then won't Discord transform it into a mention like it does when typing `@skyra` in SDS?

src/lib/i18n/languageKeys/keys/settings/channels/Logs.ts
```diff
@@ -0,0 +1,19 @@
...
+export const Member = T<string>('settings:channelsLogsMember');
+export const Message = T<string>('settings:channelsLogsMessage');
+export const Moderation = T<string>('settings:channelsLogsModeration');
+export const NsfwMessage = T<string>('settings:channelsLogsNsfwMessage');
+export const Prune = T<string>('settings:channelsLogsPrune');
+export const Reaction = T<string>('settings:channelsLogsReaction');
+export const RoleCreate = T<string>('settings:channelsLogsRoleCreate');
+export const RoleDelete = T<string>('settings:channelsLogsRoleDelete');
+export const RoleUpdate = T<string>('settings:channelsLogsRoleUpdate');
+export const ServerUpdate = T<string>('settings:channelsLogsServerUpdate');
```
favna, line 19: Bunch of these are missing from `en-US/settings.json`. They are used in `s!conf`.

favna (changes requested):
events could really use a `LogEvent` to get rid of a lot of duplicate code but it's fine to do that in another PR.

src/languages/en-US/guilds.json
kyranet: Embeds only, and no, to ping a user or a role you need to use `<@{user.id}>` or `<@&{role.id}>` respectively.

src/lib/i18n/languageKeys/keys/settings/channels/Logs.ts
favna: Note that you can get the content from skyra.pw because generally the tooltip text was made to match what's in settings. But do not my review comments on there that changes those.

kyranet: Fixed in 6c6a6a06.

src/languages/en-US/settings.json
```diff
@@ -9,13 +9,23 @@
...
 	"channelsIgnoreMessageEdit": "Channels I should ignore when checking for edited messages to log.",
 	"channelsIgnoreReactionAdd": "Channels I should ignore when checking for added reactions.",
+	"channelsLogsChannelCreate": "The channel for channel creation logs, if set, I will send a message when another channel is created.",
+	"channelsLogsChannelDelete": "The channel for channel deletion logs, if set, I will send a message to this channel when another channel is deleted.",
+	"channelsLogsChannelUpdate": "The channel for channel update logs, if set, I will send a message to this channel when any channel (including this one) gets updated in any way. This message will contain the changes made to the channel.",
+	"channelsLogsEmojiCreate": "The channel for emoji creation logs, if set, I will send a message when an emoji has been created.",
+	"channelsLogsEmojiDelete": "The channel for emoji update logs, if set, I send a message when an emoji is updated in any way. This message will contain the changes made to the emoji.",
+	"channelsLogsEmojiUpdate": "The channel for emoji deletion logs, if set, I will send a message when an emoji is deleted.",
```
favna:
```suggestion
	"channelsLogsEmojiDelete": "The channel for emoji deletion logs, if set, I will send a message when an emoji is deleted.",
	"channelsLogsEmojiUpdate": "The channel for emoji update logs, if set, I send a message when an emoji is updated in any way. This message will contain the changes made to the emoji.",
```

flipped

favna (changes requested):

favna (approved):

PR #1736, opened 2021-04-06
feat(fun): added pop

kyranet:


src/lib/util/util.ts
```diff
@@ -682,7 +682,12 @@ export const shuffle = <T>(array: T[]): T[] => {
 	return array;
 };
 
-export const random = (num: number) => Math.round(Math.random() * num);
+export const random = (num: number) => Math.floor(Math.random() * num);
+
+export const randomBetween = (a: number, b: number) => {
+	const [minimum, maximum] = a > b ? [b, a] : [a, b];
+	return random(maximum - minimum) + minimum;
+};
```
favna: test coverage goes ⬇️ 

That is to say, We have a test for `random` wherein `Math.random` is mocked. So this should be tested also.

favna (changes requested):

src/lib/util/util.ts
kyranet: That's left-over from earlier iterations.

favna (approved):

PR #1739, opened 2021-04-06
feat: allow all text commands to run in news too

kyranet:


favna (approved):
Wouldn't be much easier to just modify SkyraCommand to implicitly add news when setting text? I mean it LGTM but...

PR #1740, opened 2021-04-06
fix: twitch fixes

favna:


src/commands/Twitch/twitchsubscription.ts
```diff
@@ -45,14 +45,13 @@ export class UserCommand extends SkyraCommand {
...
 		const status = await args.pick(UserCommand.status);
-		const content = await args.rest('string').catch(() => null);
 		const entry: Entry = {
 			author: message.author.id,
 			channel: channel.id,
 			createdAt: Date.now(),
 			gamesBlacklist: [],
 			gamesWhitelist: [],
-			message: content ?? null,
+			message: null,
```
kyranet: Uh?

Favna: We haven't supported message for add since the switch to sapphire. This was still a leftover from when we didn't have subcommand specific arg parsing.

kyranet (approved):

PR #1748, opened 2021-04-07
refactor: move word scan to worker

kyranet:
- [x] Send RegExp on each call instead of pre-caching it in many workers.
- [x] Add the WorkerManager to injectedContext.
- [x] Actually use the workers.


favna (commented):
LGTM but it would be really nice if this would get tested

favna (approved):

PR #1751, opened 2021-04-08
feat: add initial role specifically for humans and bots

Stitch07:


favna (changes requested):
New keys need to be added to Skyra.Database as well. However as there are many changes to those files in another branch this PR is temporarily blocked.

favna (approved):

PR #1753, opened 2021-04-08
fix(moderation): handle payloads correctly

kyranet:
`ModerationMessageEvent#preProcess` is very lax and allows for any output besides numbers.

#1748 introduced a regression where the word handler's return type would be changed to something else, but always a non-null payload in case a RegExp existed.

During the tests, I tested every case in the worker via eval and stuff, everything worked just fine.

Except the monitor... there was no integration testing, and I tested with filtered words only, it didn't occur to me to test on non-filtered ones, and it didn't trigger on us (Alphy was running for quite some time) because the testing server was dead silent, and the ones active were moderators+, which aren't affected.

For preventive measures, I strict-typed the returns of all moderation filters, so in case something changes, TypeScript will complain.

This issue was so silly, it makes me want to go to a corner and cry for a day.

I will try to add integration testing for moderation monitors later, mocking the components and making them go through a very high flow of messages, as the server spawned 6 workers in 25 seconds, which should not have happened.

I am deeply sorry for this regression. 😭

### DIFF
--- a/src/events/moderation/messages/guildUserMessageAttachmentsHandler.ts
+++ b/src/events/moderation/messages/guildUserMessageAttachmentsHandler.ts
@@ -21,7 +21,7 @@ import type { TFunction } from 'i18next';
 	}
 })
 export class UserModerationMessageEvent extends ModerationMessageEvent {
-	protected preProcess(message: GuildMessage) {
+	protected preProcess(message: GuildMessage): 1 | null {
 		const attachments = message.attachments.size;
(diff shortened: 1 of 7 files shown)


vladfrangu:
> This issue was so silly, it makes me want to go to a corner and cry for a day.

Sure hope that isn't the case. Things like these happen.

vladfrangu (approved):

favna (approved):

PR #1789, opened 2021-04-20
feat: dockerize skyra

favna:
<details>
<summary>Draft description</summary>

TODO:
- [x] Change to multi-stage build container instead of building TS code outside of container
- [x] ~~Change building of Skyra to ESBuild~~ ESBuild wont work for Skyra until we go full ESM
- [x] Write README specific for Skyra's dockerhub page
- [x] Test API
- [x] Test WebSockets
- [x] Change continuous delivery target branch back to main
- [x] Profile and canvas fonts
- [x] Test everything on a Linux machine
    - ~~I asked Nyte if he can assist us here, he is willing to help.~~
    - ~~In particular `lavalink` **has to** run on `network mode` `host` for ipv6 on the VPS. This means that we need the Linux equivalent of `host.docker.internal`~~
    - See https://github.com/docker/for-linux/issues/264#issuecomment-823528103
    - Tested using a simple alpine container and calling `curl host.docker.internal:2333/decode` to confirm connection to lavalink is possible.
    - Tested connecting to other containers the same with but with `curl http://hasteserver:8090/raw/about` to get the https://hastebin.skyra.pw/raw/about document.

</details>


typescript/src/lib/SkyraClient.ts
```diff
@@ -99,9 +99,11 @@ export class SkyraClient extends SapphireClient {
 
 		if (envIsDefined('REDIS_AFK_DB') && envParseBoolean('REDIS_ENABLED')) {
 			Store.injectedContext.afk = new Redis({
+				host: envParseString('REDIS_HOST'),
 				port: envParseInteger('REDIS_PORT'),
 				db: envParseInteger('REDIS_AFK_DB'),
-				password: envParseString('REDIS_PASSWORD')
+				password: envParseString('REDIS_PASSWORD'),
+				lazyConnect: true
```
favna, line 106: fyi @kyranet this line is a bugfix of sorts because rn it tries to connect to reddis twice at the same time. One would think that doesn't matter shit because they're different databases, but it does cause a non-fatal, non-erroring error to be logged (it should really be a warning). 

Anyway, `lazyConnect` is:

> By default, When a new `Redis` instance is created, it will connect to Redis server automatically. If you want to keep the instance disconnected until a command is called, you can pass the `lazyConnect` option to the constructor.

https://github.com/luin/ioredis/blob/88b8a3125cd039afdac928c529a1b9958009979c/API.md#new-redisport-host-options

So it also has the added benefit that until something AFK is called, the connection is not fully established.

kyranet (approved):

Tylertron1998 (approved):

PR #1801, opened 2021-04-23
feat: added AFK command and handler

kyranet:


typescript/src/events/guilds/messages/guildUserMessageAfkHandler.ts
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	private async removeAfk(message: GuildMessage) {
+		// Get the entry, if any:
+		const entry = await this.getAfk(message.guild.id, message.author.id);
+
+		// If it does not exist, do nothing:
+		if (entry === null) return;
+
+		// If the message was sent within a minute, do nothing:
```
favna:
```suggestion
		// If the message was sent within 30 seconds, do nothing:
```

`this.threshold` is 30 seconds but the comment says 1 minute

favna (changes requested):

favna (approved):

PR #1818, opened 2021-05-01
refactor: switch to @sapphire/fetch

favna:
sapphire packages go brrrr


kyranet (commented):
~~CI's red due to compiler errors in tests.~~
~~Note: if Jest is going to continue being terrible, we might as well switch to another or write our own.~~

Edit from favna: @kyranet is full of bullshit. It was not Jest's fault. As proven by #1819

kyranet (approved):

PR #1822, opened 2021-05-02
feat(join-and-leave): add auto-delete timer options

kyranet:


typescript/src/languages/en-US/settings.json
```diff
@@ -44,7 +44,9 @@
 	"language": "The language I will use for your server. It may not be available in the language you want.",
 	"messagesAnnouncementEmbed": "Whether to send announcements in embeds or not",
 	"messagesFarewell": "The message I shall send to when a user leaves. You must set up `channels.farewell` and `events.member-remove`",
+	"messagesFarewellAutoDelete": "The amount of time before deleting a farewell message after it has been sent. If not set, the message will not be deleted.",
```
favna, line 47: We should remember if this key shows up in crowdin after a bit after merging because we disabled push sources

favna (approved):

PR #1844, opened 2021-05-09
fix: resolve proper case id

favna:


typescript/src/commands/Suggestions/suggest.ts
```diff
@@ -106,7 +106,7 @@ export class UserCommand extends SkyraCommand {
 		const [{ max }] = (await suggestions.query(
 			/* sql */ `
 			SELECT max(id)
-			FROM suggestion
+			FROM ${suggestions.metadata.tableName}
```
kyranet:
```suggestion
			FROM "${suggestions.metadata.tableName}"
```

typescript/src/lib/moderation/managers/ModerationManager.ts
```diff
@@ -122,12 +122,21 @@ export class ModerationManager extends Collection<number, ModerationEntity> {
 		return this;
 	}
 
-	public async count() {
-		if (this._count === null) {
-			const { moderations } = this.db;
-			this._count = await moderations.count({ where: { guildID: this.guild.id } });
-		}
-		return this._count!;
+	public async getCurrentModerationID() {
```
kyranet: No need to prefix ID as a moderation one, inside a moderation manager.
```suggestion
	public async getCurrentID() {
```

kyranet: Also rename all instances calling this.

typescript/src/lib/moderation/managers/ModerationManager.ts
```diff
@@ -122,12 +122,21 @@ export class ModerationManager extends Collection<number, ModerationEntity> {
...
+		const { moderations } = this.db;
+
+		const [{ max }] = (await moderations.query(
+			/* sql */ `
+				SELECT max(case_id)
+				FROM ${moderations.metadata.tableName}
+				WHERE guild_id = $1
+				`,
+			[this.guild.id]
+		)) as [MaxQuery];
```
kyranet: This should be called only if `this._count` is null.

typescript/src/lib/moderation/managers/ModerationManager.ts
```diff
@@ -122,12 +122,21 @@ export class ModerationManager extends Collection<number, ModerationEntity> {
...
-			this._count = await moderations.count({ where: { guildID: this.guild.id } });
-		}
-		return this._count!;
+	public async getCurrentModerationID() {
+		const { moderations } = this.db;
+
+		const [{ max }] = (await moderations.query(
+			/* sql */ `
+				SELECT max(case_id)
+				FROM ${moderations.metadata.tableName}
```
kyranet:
```suggestion
				FROM "${moderations.metadata.tableName}"
```

kyranet (changes requested):

kyranet (approved):

PR #1845, opened 2021-05-09
chore: remove update command

favna:


docker/docker-compose.yml
```diff
@@ -137,6 +137,7 @@ services:
     networks:
       - skyra
     ports:
+      - '8282:8282'
```
kyranet, line 140: What's that port?

kyranet (approved):

PR #1847, opened 2021-05-12
feat(fun): added think command

kyranet:


typescript/src/languages/en-US/commands/fun.json
```diff
@@ -193,6 +193,18 @@
 			"Skyra"
 		]
 	},
+	"thinkDescription": "Think, Mark!",
+	"thinkExtended": {
+		"usages": [
+			"User"
+		],
+		"extendedHelp": "\"Think, Mark!\" is a meme from the TV show Invincible, the rest... no, that would be spoilers.",
+		"reminder": "This command contains animated blood, which may not be suitable for all audiences.",
```
favna:
```suggestion
		"reminder": "This command contains cartoon blood, which may not be suitable for all audiences.",
```

animated in this context implies movement, but it's just a png.

favna (changes requested):

favna (approved):

PR #1877, opened 2021-05-29
refactor: use UserError where applicable

favna:


typescript/src/commands/Misc/afk.ts
```diff
@@ -51,8 +51,8 @@ export class UserCommand extends SkyraCommand {
 		return allResult.success ? this.clearAll(message, args) : this.clearUser(message, args);
 	}
 
-	@requiresLevel(PermissionLevels.Moderator, async (_: GuildMessage, args: SkyraCommand.Args) => {
-		throw args.t(LanguageKeys.Commands.Misc.AfkPermissionLevelList);
+	@requiresLevel(PermissionLevels.Moderator, async () => {
+		throw new UserError({ identifier: LanguageKeys.Commands.Misc.AfkPermissionLevelList });
 	})
```
kyranet: Have you thought of... making the decorators optionally accept a string and wrap them into this boiler-plate, or simply making a wrapper function, so we don't need to import UserError?

kyranet (approved):

PR #1899, opened 2021-06-19
refactor: migrate to updated @sapphire/utilities & bump CI to Node 16

favna:


typescript/src/commands/Management/Message Filters/messageMode.ts
```diff
@@ -1,8 +1,8 @@
 import { AdderKey, GuildEntity, GuildSettings } from '#lib/database';
 import { LanguageKeys } from '#lib/i18n/languageKeys';
 import { SelfModerationCommand } from '#lib/moderation';
-import type { KeyOfType } from '#lib/types';
 import { ApplyOptions } from '@sapphire/decorators';
+import { PickByValue } from '@sapphire/utilities';
```
kyranet: Shouldn't this be `import type`?

kyranet: Same goes for every other instance of this.

kyranet (changes requested):

kyranet (approved):

PR #1935, opened 2021-07-08
fix(logging): fixed failed task logging

### LINKED ISSUE #1934: bug: TaskError method arguments are incorrect.
**Describe the bug**

After writing a scheduled task, I noticed I was getting useless error output, after some time debugging I found that the arguments are being passed to TaskError in the wrong order. This causes TaskError its self to throw an error leading to misleading and unhelpful error messages after a ScheduledTask has thrown an error.

The TaskError run method expects `error: Error, context: TaskErrorPayload` parameters, but whats actually being passed are `entity: ScheduleEntity, task: UserTask, error: Error`

**To Reproduce**

1. Write a Scheduled task that will throw an error
2. Run the bot until the Scheduled task runs
3. See error

**Expected behavior**
The original Error from the Scheduled task should be output

**Screenshots**

**Additional context**

favna:
This will at least help us find issues in schedules

This also fixes task errors being sent to Sentry properly.

Tested by forcing an error in `reminder.ts`:

Fixes #1934

### DIFF
--- a/typescript/src/lib/types/Augments.d.ts
+++ b/typescript/src/lib/types/Augments.d.ts
@@ -154,8 +154,8 @@ declare module '@sapphire/framework' {
 		emit(event: Events.MusicConnect, queue: Queue, voiceChannelID: string): boolean;
 		emit(event: Events.ResourceAnalyticsSync): boolean;
 		emit(event: Events.TwitchStreamHookedAnalytics, status: TwitchStreamStatus): boolean;
-		emit(event: string | symbol, ...args: any[]): boolean;
 		emit(event: Events.TaskError, error: Error, payload: TaskErrorPayload): boolean;
+		emit(event: string | symbol, ...args: any[]): boolean;
 	}
 }
 
--- a/typescript/src/lib/types/Augments.d.ts
+++ b/typescript/src/lib/types/Augments.d.ts
@@ -27,7 +27,8 @@ import type {
 	SplitOptions,
 	StringResolvable,
 	TextChannel,
-	User
+	User,
+	VoiceChannel
 } from 'discord.js';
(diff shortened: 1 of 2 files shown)


typescript/src/lib/types/Augments.d.ts
```diff
@@ -153,6 +111,52 @@ declare module '@sapphire/framework' {
...
+		emit(event: Events.MusicSongSkipNotify, channel: MessageAcknowledgeable, entry: QueueEntry): boolean;
+		emit(event: Events.MusicSongVolumeUpdate, queue: Queue, next: number): boolean;
+		emit(event: Events.MusicSongVolumeUpdateNotify, channel: MessageAcknowledgeable, previous: number, next: number): boolean;
+		emit(event: Events.MusicVoiceChannelJoin, queue: Queue, voiceChannel: VoiceChannel): boolean;
+		emit(event: Events.MusicVoiceChannelLeave, queue: Queue): boolean;
+		emit(event: Events.MusicConnect, queue: Queue, voiceChannelID: string): boolean;
+		emit(event: Events.ResourceAnalyticsSync): boolean;
+		emit(event: Events.TwitchStreamHookedAnalytics, status: TwitchStreamStatus): boolean;
+		emit(event: string | symbol, ...args: any[]): boolean;
+		emit(event: Events.TaskError, error: Error, payload: TaskErrorPayload): boolean;
```
kyranet:
```suggestion
		emit(event: Events.TaskError, error: Error, payload: TaskErrorPayload): boolean;
		emit(event: string | symbol, ...args: any[]): boolean;
```

kyranet (changes requested):

kyranet (approved):

Issue #1833, opened 2021-05-03
request: streamline viewing moderation logs for a specific user


favna:
**Is your feature request related to a problem? Please describe.**

Right now we have 4 commands: `warnings`, `mutes`, `history` and `moderations`, and the former 2 of these actually call methods in `moderations`. This is extremely confusing for end-users. For example, when they just want to view all the actions taken on a specific they'd have to write `s!history details User` which really makes no sense at all.

**Describe the solution you'd like**

We should merge these 4 commands into just 1 command which will be `moderations` with aliases of `warnings`, `mutes`, `mod-logs` and `mod-log`. This command will always show:
- All mutes, warnings, bans, kicks, etc for a user
- Including permanent and temporary ones. 
  - Non-expired temporary ones show when they will expire like so: Expires in: 1 minute
  - Expired temporary ones will have a strikethrough effect like so: Expired
  - Expired ones will have have the "Temporary <Action>" striked through, as well as their reason if any.
  - There will be flags (`--no-strikethrough`, `--no-strike`) to disable the strikethrough effect. This is just in case people want to read it without the strikethrough styling as strikethrough does make the text slightly harder to read.
- The UPM will have a footer with a counter of how many warnings, mutes, kicks and bans the user has similar to `history overview User`

This means there is no more confusion between details and overview, and all data is always shown

**Describe alternatives you've considered**

N.A. We need to provide better UX when it comes to these commands.

**Additional context**

Mockup:

![image](https://user-images.githubusercontent.com/4019718/116926442-a62e0500-ac5a-11eb-8b44-44bc33e62334.png)

PR #1942, opened 2021-07-12
feat(history): add subcommand aliases

kyranet:


Favna:
Look I hate to do it to ya man but I have to reject this because the issue for these changes is https://github.com/skyra-project/skyra/issues/1833 and this does not make those changes at all. It's pretty much useless because of the changes requested in that issue.

favna (approved):

PR #1945, opened 2021-07-12
feat(define): add `def` alias

kyranet:


typescript/src/commands/Tools/Dictionary/define.ts
```diff
@@ -12,7 +12,7 @@ import type { TFunction } from 'i18next';
 
 @ApplyOptions<PaginatedMessageCommand.Options>({
 	enabled: envIsDefined('OWLBOT_TOKEN'),
-	aliases: ['definition', 'defination', 'dictionary'],
+	aliases: ['def', 'definition', 'dictionary'],
```
favna:
```suggestion
	aliases: ['def', 'definition', 'defination', 'dictionary'],
```

The misspelling was there intentionally because it's one of those words that people often misspell.

favna (changes requested):

favna (approved):

PR #1949, opened 2021-07-16
feat(social): add `reset --all`

kyranet:


typescript/src/commands/Levelling/Management/social.ts
```diff
@@ -91,9 +92,25 @@ export class UserCommand extends SkyraCommand {
...
+		if (confirmed === null) this.error(LanguageKeys.Commands.Social.SocialResetAllTimeOut);
+		if (!confirmed) this.error(LanguageKeys.Commands.Social.SocialResetAllAborted);
+
+		const { members } = this.context.db;
+		const result = await members.delete({ guildID: message.guild.id });
+		if (!result.affected) this.error(LanguageKeys.Commands.Social.SocialResetAllEmpty);
+
+		// Delete the local leaderboard entry since it's all set to 0 at this point.
+		this.context.client.leaderboard.local.delete(message.guild.id);
+		return message.send(args.t(LanguageKeys.Commands.Social.SocialResetAllSuccess, { count: result.affected }));
```
favna, line 114:
```suggestion
		return message.channel.send(args.t(LanguageKeys.Commands.Social.SocialResetAllSuccess, { count: result.affected }));
```

DJS v13 preps...

favna (changes requested):

typescript/src/commands/Levelling/Management/social.ts
kyranet: No.

We're not removing editable and deletable commands, we'll switch those to functions in a new major version of the editable commands package, which I'll start working on tomorrow.

favna (approved):

PR #1965, opened 2021-07-19
refactor(permissions): switch to functional

kyranet:


typescript/src/preconditions/Spam.ts
```diff
@@ -1,20 +1,21 @@
...
+import { isModerator, isOwner } from '#utils/functions';
 import { AsyncPreconditionResult, Precondition } from '@sapphire/framework';
 import { RateLimitManager } from '@sapphire/ratelimits';
 import { Time } from '@sapphire/time-utilities';
-import type { Message } from 'discord.js';
 
 export class UserPrecondition extends Precondition {
 	private readonly ratelimit = new RateLimitManager(Time.Second * 30, 1);
 
-	public async run(message: Message): AsyncPreconditionResult {
```
favna, line 11: Big no no. We use this precondition in DM capable commands as well. Just search for `spam: true`. For example catfact and dog.

favna (changes requested):

typescript/src/preconditions/Spam.ts
kyranet: Look at literally the line below

favna (approved):

PR #1967, opened 2021-07-20
refactor(messages): switch to functional

kyranet:
50% completed? :P


typescript/src/lib/util/functions/channels.ts
```diff
@@ -0,0 +1,24 @@
+import { Permissions, GuildChannel, Message } from 'discord.js';
+
+export type ChannelTypes = Message['channel'];
+export type GuildBasedChannelTypes = Extract<ChannelTypes, GuildChannel>;
```
favna, line 4: Are these exported types really necessary? I can smell them being duplicated in various places coming.

favna (approved):

PR #1969, opened 2021-07-21
refactor: move utilities to specific files

kyranet:
Also renamed `wrap` to `safeWrapPromise`


typescript/src/lib/SkyraClient.ts
```diff
@@ -60,7 +60,7 @@ export class SkyraClient extends SapphireClient {
 	public readonly analytics: AnalyticsData | null;
 
 	@enumerable(false)
-	public readonly guildMemberFetchQueue: GuildMemberFetchQueue = new GuildMemberFetchQueue(this);
+	public readonly guildMemberFetchQueue: GuildMemberFetchQueue = new GuildMemberFetchQueue();
```
favna, line 63: is this a bugfix 🤔 ?

kyranet: Nah, we already have access to `client` via container, so I refactored it.

favna (dismissed):

favna (approved):

PR #1984, opened 2021-07-25
refactor: remove even more structures

kyranet:
Only extensions left are the SkyraGuild ones, which are far more difficult to port and change a lot more code due to their cache.

### DIFF
--- a/typescript/src/lib/util/functions/messages.ts
+++ b/typescript/src/lib/util/functions/messages.ts
@@ -163,3 +163,12 @@ export async function promptConfirmation(message: Message, options: string | Pro
 	const response = (await message.send(options)) as Message;
 	return canReact(response) ? promptConfirmationReaction(message, response, options) : promptConfirmationMessage(message, response, options);
 }
+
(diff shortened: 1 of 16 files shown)


typescript/src/lib/util/functions/messages.ts
```diff
@@ -163,3 +163,12 @@ export async function promptConfirmation(message: Message, options: string | Pro
 	const response = (await message.send(options)) as Message;
 	return canReact(response) ? promptConfirmationReaction(message, response, options) : promptConfirmationMessage(message, response, options);
 }
+
+export async function promptForMessage(message: Message, sendOptions: string | MessageOptions, time = Time.Minute): Promise<string | null> {
+	// The cast is for mismatching overloads:
+	const response = (await message.channel.send(sendOptions as any)) as Message;
```
favna, line 170:
```suggestion
	// TODO: v13 | Remove typecast
	const response = (await message.channel.send(sendOptions as any)) as Message;
```

So we can find it back easily with [ToDo Tree](https://marketplace.visualstudio.com/items?itemName=Gruntfuggly.todo-tree)

favna (changes requested):

favna (approved):

PR #1989, opened 2021-07-26
feat(animation): add anilist

favna:
⪢ The old `anime` is now `kitsu-anime`
⪢ The old `manga` command is now `kitsu-manga`

- Changed all the `anime.json` to `animation.json` to match the TS file for these keys
- Sorted `animation.json` keys
- Prefixed old keys of manga and anime with `kitsu`
- Sorry that Git doesn't detect them as moved files :(
- Add new anime command


src/commands/Animation/anime.ts
```diff
@@ -1,122 +1,134 @@
...
-	extendedHelp: LanguageKeys.Commands.Animation.AnimeExtended
+	description: LanguageKeys.Commands.Animation.AniListAnimeDescription,
+	extendedHelp: LanguageKeys.Commands.Animation.AniListAnimeExtended
 })
 export class UserPaginatedMessageCommand extends PaginatedMessageCommand {
 	public async run(message: GuildMessage, args: PaginatedMessageCommand.Args) {
-		const animeName = await args.rest('string');
-		const response = await sendLoadingMessage(message, args.t);
+		const { t } = args;
+		const [search, loadingMessage] = await Promise.all([args.rest('string'), sendLoadingMessage(message, t)]);
```
kyranet: You seriously have no idea how much I hate this.

`args.rest('string')` is basically zero cost, it takes sub-ms to run, but regardless of whether it succeeds or not, you're always sending a message at potentially the same time `args.rest` could send another with a rejection.

kyranet (approved):

PR #2010, opened 2021-08-18
feat: update twitch to eventsub

favna:
eventsub is so much better
new db structure is so much better
 
twitch stuff is now so much better

Priority critical because webhook dies in September.

**Tasks:**

- [x] Testing migration
- [x] Testing with actual events being registered through my skyrahost.xyz
- [x] Remove TODO in Twitch.ts
- [x] Remove `fillData.mjs` and `FillData.sql`


src/routes/twitch/twitchEventSubVerify.ts
```diff
@@ -0,0 +1,76 @@
...
+	}
+}
+
+declare module 'http' {
+	interface IncomingHttpHeaders extends NodeJS.Dict<string | string[]> {
+		'twitch-eventsub-message-signature'?: string;
+		'twitch-eventsub-message-id'?: string;
+		'twitch-eventsub-message-timestamp'?: string;
+	}
+}
```
kyranet: Do we really need this work-around? It affects all instances of `request.headers`.

src/commands/Twitch/twitchsubscription.ts
```diff
@@ -41,269 +25,313 @@ type Entry = NotificationsStreamsTwitchStreamer;
...
-				// Insert the entry to the database performing an upsert, if it created the entry, we tell the Twitch manager
-				// to send Twitch a message saying "hey, I want to be notified, can you pass me some data please?"
-				if (await this.upsertSubscription(message.guild, streamer)) {
-					await this.container.client.twitch.subscriptionsStreamHandle(streamer.id, TwitchHooksAction.Subscribe);
-				}
-			} else {
-				// Retrieve the subscription.
-				const raw = settings[this.#kSettingsKey][subscriptionIndex];
+		const subscriptionType = await args.pick(UserCommand.status);
+		const customMessage = await args.rest('string', { maximum: 50 }).catch(() => null);
```
kyranet: Isn't 50 obnoxiously small?

src/commands/Twitch/twitchsubscription.ts
```diff
@@ -41,269 +25,313 @@ type Entry = NotificationsStreamsTwitchStreamer;
...
+			// Retrieve the current highest ID in the database, TypeORM doesn't resolve this automatically for some reason
+			const [{ max }] = (await twitchSubscriptions.query(/* sql */ `
+				SELECT max(id)
+				FROM public.twitch_subscriptions;
+			`)) as [MaxQuery];
+
+			// If the max value is not nullish or zero then set it in the entity instance
+			if (!isNullishOrZero(max)) {
+				twitchSubscriptionEntity.id = max + 1;
+			}
```
kyranet: What is this prone-to-error heresy?

src/commands/Twitch/twitchsubscription.ts
```diff
@@ -41,269 +25,313 @@ type Entry = NotificationsStreamsTwitchStreamer;
...
-				await em.remove(toDelete);
-				await em.save(toUpdate);
+			this.error(LanguageKeys.Commands.Twitch.TwitchSubscriptionRemoveStreamerStatusNotMatch, {
+				streamer: streamer.display_name,
+				status: this.getSubscriptionStatus(subscriptionType, statuses)
 			});
+		}
+
+		// Get all subscriptions for this streamer, status and channel
+		const streamerWithStatusHasChannel = statuses.filter((guildSubscription) => guildSubscription.channelId === channel.id);
```
kyranet: We're only removing the first entry, why is this `filter` and not `find`?

src/commands/Twitch/twitchsubscription.ts
```diff
@@ -41,269 +25,313 @@ type Entry = NotificationsStreamsTwitchStreamer;
...
+		const { twitchSubscriptions } = this.container.db;
+
+		// Get all subscriptions for the provided streamer and subscription type
+		const subscription = await twitchSubscriptions.findOne({
+			relations: ['guildSubscription'],
+			where: {
+				streamerId: streamer.id,
+				subscriptionType
+			}
+		});
```
kyranet: When we delete the guild subscription, we have `subscriptionId`, which we can use alongside a `LIMIT 1` to not do a very costly `JOIN LEFT` every time a subscription is removed.

kyranet (changes requested):

src/routes/twitch/twitchEventSubVerify.ts
favna: Without the type of the headers are `string | string[]` which means type casting elsewhere
image

src/commands/Twitch/twitchsubscription.ts
favna: I've explained this many times already. If you do not understand the code then ask it in Discord instead of leaving review comments like this.

First of all, `findOne` already is `LIMIT 1`. It is literally in the name

Secondly, no we DO need to have `relations: ['guildSubscription']`. I've explained this a million times already now. This function is to ensure that once the last guild unsubs from a streamer, we send the unsub request to Twitch. So we DO need to know which guilds are subbed with a LEFT JOIN here. And it works because we first `await` the `GuildSubscriptionEntity#remove` just above where this function gets called. So at that point if this returns 0 `guildSubscription` we know we can tell Twitch to unsub.

Thirdly, sure I'll change it to matching on subscriptionId but that doesn't change point 2 at all.

src/commands/Twitch/twitchsubscription.ts
favna: This is not prone to error. It is literally the same code we use for moderation cases and suggestion ids. In fact it is YOUR code for those parts of Skyra.

src/routes/twitch/twitchEventSubVerify.ts
kyranet: You can cast them elsewhere, this is similar to defining global variables for a single place.

src/commands/Twitch/twitchsubscription.ts
kyranet: > It is literally the same code we use for moderation cases and suggestion ids.

No, it's not.

Moderation cases and suggestion IDs increment by one **based** on `guild_id`, which is why it needs to be fetched manually.

Twitch subscriptions are incremented per table, they can use PGSQL's internal counter, which is both atomic and automatic.

We can't use that counter in the other cases because PGSQL doesn't have a per-value counter.

ref:
- [`SERIAL` docs](https://www.tutorialspoint.com/postgresql/postgresql_using_autoincrement.htm)
- [TypeORM docs](https://typeorm.io/#/decorator-reference/primarygeneratedcolumn)

kyranet (approved):

PR #2033, opened 2021-09-13
fix(api): fixed thread channels in ApiTransformer

favna:


src/lib/api/ApiTransformers.ts
```diff
@@ -208,6 +222,21 @@ function flattenChannelDM(channel: DMChannel): FlattenedDMChannel {
...
+function flattenChannelThread(channel: ThreadChannel): FlattenedThreadChannel {
+	return {
+		id: channel.id,
+		type: channel.type,
+		archived: channel.archived ?? false,
+		archivedTimestamp: channel.archiveTimestamp,
+		createdTimestamp: channel.createdTimestamp,
+		guildId: channel.guildId,
+		name: channel.name,
+		parent: channel.parent ? flattenChannel(channel.parent) : null,
```
kyranet: Please don't do this, just parentId is enough.

kyranet (changes requested):

kyranet (approved):

PR #2097, opened 2021-10-17
fix(lockdown): ensure proper permission checks

favna:
Ref for inspiration: https://github.com/Naval-Base/yuudachi/pull/1012

Yuu's post-merge files: https://github.com/Naval-Base/yuudachi/tree/main/src/functions/lockdowns and https://github.com/Naval-Base/yuudachi/blob/main/src/commands/moderation/lockdown.ts

### DIFF
--- a/src/commands/Moderation/lockdown.ts
+++ b/src/commands/Moderation/lockdown.ts
@@ -5,12 +5,13 @@ import { PermissionLevels } from '#lib/types/Enums';
 import { floatPromise } from '#utils/common';
 import { assertNonThread, getSecurity } from '#utils/functions';
 import { clearAccurateTimeout, setAccurateTimeout } from '#utils/Timers';
+import { channelMention } from '@discordjs/builders';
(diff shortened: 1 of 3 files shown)


src/commands/Moderation/lockdown.ts
```diff
@@ -61,16 +72,34 @@ export class UserCommand extends SkyraCommand {
...
+				allow: 0n,
+				deny: this.permissionsToDeny,
+				type: 'role'
+			},
+			{
+				id: this.container.client.user!.id,
+				allow: this.clientPermissionsToAllow,
+				type: 'member'
+			}
+		]);
```
kyranet: This... is very troublesome. It should account for existent permissions, otherwise it'll run the risk of removing other permissions such as restrictions or overrides for other roles (e.g. mods) to even send messages.

kyranet (changes requested):

kyranet:
Superseded by #2593

PR #2122, opened 2021-11-13
fix: edit namespace linting

Lioness100:
`@typescript-eslint/no-namespace` was set to `"off"` in the latest `@sapphire/eslint-config` version, so there's no more need for config overrides or disabling comments.

I don't think namespaces were even used in src/lib/util/constants.ts, but 🤷


kyranet (dismissed):
Thanks!

favna (approved):

PR #2156, opened 2021-12-28
fix(heapsnapshot): fixed it for docker

favna:


src/commands/System/Admin/heapsnapshot.ts
```diff
@@ -38,11 +43,24 @@ import { writeHeapSnapshot } from 'node:v8';
...
 
-		return send(message, `Captured in \`${filename}\`, check! Remember, do NOT share this with anybody, it may contain a lot of sensitive data.`);
+		const outputMessage = [
+			//
+			`Heapsnapshot stored at ${inlineCodeBlock(fileName)}!`,
+			envParseString('NODE_ENV') === 'development'
+				? `\nThis file can be extracted from the Docker container using the following command: ${inlineCodeBlock(
+						`docker cp skyra:${join(rootFolder, fileName)} ./${fileName}`
+				  )}`
+				: null,
```
kyranet: ... only in development?

kyranet (changes requested):

src/commands/System/Admin/heapsnapshot.ts
favna: woops

kyranet (approved):

PR #2167, opened 2022-01-03
fix: resolved message-update logs not showing channel

kyranet:


favna (commented):
Shouldn't we use `channelMention` from builders and pass a string value to i18next?

kyranet:
No, because embed footers don't render mentions.

favna (dismissed):
oh right shit

favna (approved):

PR #2187, opened 2022-02-03
fix(birthday): updated how the age is calculated

### LINKED ISSUE #2186: bug: Wrong age in Birthday message with {age}
**Describe the bug**

So, today is my birthday, and i had configured it on the bot previously, today the message the bot sent had the wrong age in it,
my birthday is 02/03/1998, im now 24 yo, but the message the bot sent was saying that i was 23 yo.

**To Reproduce**

1. Configure your birthday
2. Configure a bot message for birthday using {age}
3. Wait for your birthday

**Expected behavior**

When using {age} in the message, i expected that i would be replaced with the new age (in my case, 24, not 23).

**Screenshots**

image
image

**Additional context**

Since i had free time, i decided to explore the code, after understanding how getAge function works, I then went to the tasks, and saw this:

```
// tasks/birthday.ts Lines 96-102

        if (canSendMessages(channel)) {
            await channel.send({
                content: this.transformMessage(content, member.user, getAge(data), t),
                allowedMentions: { users: [member.id] }
            });
        }
```

Sorry if im wrong, but shouldn't `getAge(data)` be add +1 ? since it will return the old age if the day and month are the same.

christianfds:
Fixes #2186

### DIFF
--- a/src/lib/birthday/utils.ts
+++ b/src/lib/birthday/utils.ts
@@ -111,13 +111,11 @@ export function compareDate(month: Month, day: number, { now = Date.now() }: Tim
 export function getAge(data: DateWithOptionalYear, { now = Date.now() }: TimeOptions = {}) {
 	if (data.year === null) return null;
 
-	// If the birthday has happened, we subtract the years by one, that way:
-	//
-	// * 2021/03/18 - 2020/05/10 = 0
-	// * 2021/03/18 - 2020/02/26 = 1
-	const extra = compareDate(data.month, data.day, { now }) === -1 ? 0 : -1;
-	const years = new Date(now).getUTCFullYear() - data.year;
-	return years + extra;
+	const nowAsDate = new Date(now);
+	const dataAsDate = new Date(data.year, data.month - 1, data.day);
+
+	const years = new Date(nowAsDate - dataAsDate).getFullYear() - new Date(0).getFullYear() - 1;
+	return years;
 }
 
 /**
--- a/src/lib/birthday/utils.ts
+++ b/src/lib/birthday/utils.ts
@@ -104,18 +104,27 @@ export function compareDate(month: Month, day: number, { now = Date.now() }: Tim
 
 /**
  * Gets the current age from a date.
- * @param data The data to compare.
+ * @param date The date to compare.
  * @param options The options for the operation of this function.
- * @returns `null` if `data.year` is `null`, a number of years otherwise.
+ * @returns `null` if `date.year` is `null`, a number of years otherwise.
  */
-export function getAge(data: DateWithOptionalYear, { now = Date.now() }: TimeOptions = {}) {
-	if (data.year === null) return null;
+export function getAge(date: DateWithOptionalYear, { now = Date.now() }: TimeOptions = {}) {
+	if (date.year === null) return null;
 
-	const nowAsDate = new Date(now);
-	const dataAsDate = new Date(data.year, data.month - 1, data.day);
+	const currentDate = new Date(now);
+	const dateOfBirth = new Date(date.year, date.month - 1, date.day);
 
-	const years = new Date(Number(nowAsDate) - Number(dataAsDate)).getFullYear() - new Date(0).getFullYear() - 1;
-	return years;
+	const diffYears = currentDate.getUTCFullYear() - dateOfBirth.getUTCFullYear();
+	const diffMonths = currentDate.getUTCMonth() - dateOfBirth.getUTCMonth();
+	const diffDays = currentDate.getUTCDate() - dateOfBirth.getUTCDate();
+
+	let age = diffYears;
+
+	if (diffMonths < 0 || (diffMonths === 0 && diffDays < 0)) {
+		age--;
+	}
+
+	return age;
 }
 
 /**
--- a/src/lib/birthday/utils.ts
+++ b/src/lib/birthday/utils.ts
@@ -111,9 +111,9 @@ export function compareDate(month: Month, day: number, { now = Date.now() }: Tim
 export function getAge(date: DateWithOptionalYear, { now = Date.now() }: TimeOptions = {}) {
 	if (date.year === null) return null;
 
-	const birthDate = new Date(date.year, date.month - 1, date.day);
+	const birthDate = Date.UTC(date.year, date.month - 1, date.day);
 
-	return new Date(now - Number(birthDate)).getUTCFullYear() - new Date(0).getUTCFullYear();
+	return new Date(now - birthDate).getUTCFullYear() - new Date(0).getUTCFullYear();
 }
 
 /**
--- a/src/lib/birthday/utils.ts
+++ b/src/lib/birthday/utils.ts
@@ -114,7 +114,7 @@ export function getAge(data: DateWithOptionalYear, { now = Date.now() }: TimeOpt
 	const nowAsDate = new Date(now);
 	const dataAsDate = new Date(data.year, data.month - 1, data.day);
 
-	const years = new Date(nowAsDate - dataAsDate).getFullYear() - new Date(0).getFullYear() - 1;
+	const years = new Date(Number(nowAsDate) - Number(dataAsDate)).getFullYear() - new Date(0).getFullYear() - 1;
 	return years;
 }
 
--- a/src/lib/birthday/utils.ts
+++ b/src/lib/birthday/utils.ts
@@ -111,20 +111,9 @@ export function compareDate(month: Month, day: number, { now = Date.now() }: Tim
 export function getAge(date: DateWithOptionalYear, { now = Date.now() }: TimeOptions = {}) {
 	if (date.year === null) return null;
 
-	const currentDate = new Date(now);
-	const dateOfBirth = new Date(date.year, date.month - 1, date.day);
+	const birthDate = new Date(date.year, date.month - 1, date.day);
 
-	const diffYears = currentDate.getUTCFullYear() - dateOfBirth.getUTCFullYear();
-	const diffMonths = currentDate.getUTCMonth() - dateOfBirth.getUTCMonth();
-	const diffDays = currentDate.getUTCDate() - dateOfBirth.getUTCDate();
-
-	let age = diffYears;
-
-	if (diffMonths < 0 || (diffMonths === 0 && diffDays < 0)) {
-		age--;
-	}
-
-	return age;
+	return new Date(now - Number(birthDate)).getUTCFullYear() - new Date(0).getUTCFullYear();
 }
 
 /**
--- a/tests/lib/Birthday.test.ts
+++ b/tests/lib/Birthday.test.ts
@@ -276,6 +276,18 @@ describe('Birthday', () => {
 			test('GIVEN day after now THEN returns 19', () => {
 				expect(getAge({ year: 2000, month: Month.May, day: 5 })).toBe(19);
 			});
+
+			test('GIVEN yesterday THEN returns 20', () => {
+				expect(getAge({ year: 2000, month: Month.April, day: 17 })).toBe(20);
+			});
+
+			test('GIVEN today THEN returns 20', () => {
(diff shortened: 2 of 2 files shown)


tests/lib/Birthday.test.ts
```diff
@@ -276,6 +276,18 @@ describe('Birthday', () => {
 			test('GIVEN day after now THEN returns 19', () => {
 				expect(getAge({ year: 2000, month: Month.May, day: 5 })).toBe(19);
 			});
+
+			test('GIVEN yesterday THEN returns 20', () => {
+				expect(getAge({ year: 2000, month: Month.April, day: 17 })).toBe(20);
+			});
+
+			test('GIVEN today THEN returns 20', () => {
+				expect(getAge({ year: 2000, month: Month.April, day: 18 })).toBe(20);
```
favna, line 285: April 18th 2000 -> April 17th 2020 should be 19
April 18th 2000 -> April 18th 2020 should be 20
April 18th 2000 -> April 19th 2020 should be 20

So the first and last test cases here are incorrect

favna (changes requested):

tests/lib/Birthday.test.ts
christianfds: Not actually, if i got it right, what we are changing here is the date of birth, not the current date. So:

Date of birth -> Current Date
April 17th 2000 -> April 18th 2020 should be 20 (birthday was yesterday)
April 18th 2000 -> April 18th 2020 should be 20 (birthday is today)
April 19th 2000 -> April 18th 2020 should be 19 (birthday is tomorrow)

christianfds: I can update the tests descriptions so it becomes more explicity (i also got a bit confused while writting it)

favna: I'm a bit confused right now as well. What I can certainly see is that where we use `getAge` in `upcoming-birthdays` command it does return the correct result right now. 

With the unchanged code it should return 26 for me. That is `February 21st 1995` -> `February 3rd 2022`. Then in `upcoming-birthday` we do `+ 1` on the age and the embed correctly shows 27 for me.

image

Diyagi: https://github.com/skyra-project/skyra/blob/54a95578d7eb5d18b4417c9c173bda5ac318842e/src/commands/Birthday/upcoming-birthdays.ts#L44-L45

So, if im not wrong, when you call `getAge`, you pass my birthdate (mine for exemple, 02/03/1998) and the optional date, which i believe is the date that the next task is schedule (mine for exemple, 02/03/2023).

With the unchanged `getAge`, if you are on the same date as you birthdate, it will still return your old age, so in the exemple above, it would return 24, then you increment +1 (`calculatedAge + 1`) which returns 25.

https://github.com/skyra-project/skyra/blob/82105e6493f005eb342f4a34ab77cc32026e0994/src/tasks/birthday.ts#L99

The problem when the bot announces someone birthday, because we are in the same date as my birthdate, the `getAge` returns my old age, and this value is not incremented like in the `upcoming-birthdays`.

src/lib/birthday/utils.ts
```diff
@@ -111,13 +111,11 @@ export function compareDate(month: Month, day: number, { now = Date.now() }: Tim
...
 
-	// If the birthday has happened, we subtract the years by one, that way:
-	//
-	// * 2021/03/18 - 2020/05/10 = 0
-	// * 2021/03/18 - 2020/02/26 = 1
-	const extra = compareDate(data.month, data.day, { now }) === -1 ? 0 : -1;
-	const years = new Date(now).getUTCFullYear() - data.year;
-	return years + extra;
+	const nowAsDate = new Date(now);
+	const dataAsDate = new Date(data.year, data.month - 1, data.day);
```
lveneris: These variables are poorly named since both were dates of some kind strictly speaking. Since `nowAsDate` is erroneous (see the following comment), remove it and ideally rename `dataAsDate` to something more fitting.

Alternatively, you could drop the intermediate variable since it adds no readability and drop the data parse into the years calculation.

src/lib/birthday/utils.ts
```diff
@@ -111,13 +111,11 @@ export function compareDate(month: Month, day: number, { now = Date.now() }: Tim
...
+
+	const years = new Date(Number(nowAsDate) - Number(dataAsDate)).getFullYear() - new Date(0).getFullYear() - 1;
```
lveneris: There is additionally no point using `new Date(now)` if you're simply going to convert it back into a number. `Date.now()` returns the same number you'd get from doing `Number(new Date(Date.now()))`, so there's no need to change `now` to a date.

Furthermore, that ternary seems to have been necessary - adding the unconditional -1 returns the wrong age for my birthday following these exact calculations, but I'm not quite sure the best way to fix that. Strangely, in my testing it works perfectly if you remove the -1 from the calculation, but that'd require further verification.

Update: It stands to reason mathematically that even on your birthday, using this standard method, subtracting 1 is never necessary. `now - birthYear = age` is always accurate when the dates are contrasted in milliseconds rather than years, which they are here. Therefore, the -1 need not be there. I verified this on both my birthday, the current date, and Fav's birthday, and all 3 returned correctly without the -1.

```suggestion
	const years = new Date(now - Number(dataAsDate)).getFullYear() - new Date(0).getFullYear();
```

lveneris (changes requested):

src/lib/birthday/utils.ts
christianfds: I changed it to getUTCFullYear and some issues raised while fixing some others, later i decided to use a "dumb" algorithm.

tests/lib/Birthday.test.ts
christianfds: I update the method with a different getAge algorithm, if you could test it i would appreciate.

christianfds: I hope that using getUTC... might fix any other issues

lveneris:
As of your latest commit, the code quality has arguably gone down significantly. This is a relatively naïve and inefficient way to compare the dates as opposed to implementing the changes we suggested.

src/lib/birthday/utils.ts
lveneris: I don't understand why when your initial solution needed a small tweak or two to be perfect.

christianfds:
I tried the suggest changes, but it didn't seem to work, the same issue kept on happening. On code quality i dont think it decreased, the code just got easier to read and understand, the code still with O(1) complexity.

By removing the -1 and changing it to use getUTCFullYear() there was a good result, but it still failed on a couple of tests.

If there is a simple change that would work, i would love to see it. Saddly i didn't manage to make it work =/

lveneris:
As seen below, having tried on my local system, my suggested changes alone and `getUTCFullYear` pass all tests - even with your new suite of them.

image

src/lib/birthday/utils.ts
```diff
@@ -104,20 +104,27 @@ export function compareDate(month: Month, day: number, { now = Date.now() }: Tim
...
+	const diffMonths = currentDate.getUTCMonth() - dateOfBirth.getUTCMonth();
+	const diffDays = currentDate.getUTCDate() - dateOfBirth.getUTCDate();
+
+	let age = diffYears;
+
+	if (diffMonths < 0 || (diffMonths === 0 && diffDays < 0)) {
+		age--;
+	}
+
+	return age;
```
lveneris:
```suggestion
	const birthDate = new Date(date.year, date.month - 1, date.day);
	
	return new Date(now - Number(birthDate)).getUTCFullYear() - new Date(0).getUTCFullYear();
```

lveneris (changes requested):

christianfds:
Thats weird, here i get two failing tests with the code suggested.

image

And those are the ones where the birthday is on the current day, i tried fiddling a bit with the time and noticed that one of the tests succeed if add 3 hours to the current date. It kinds of makes sense since im in a GMT-3 zone.
`
test('GIVEN now is birthday THEN returns 24', () => {
expect(getAge({ year: 1997, month: Month.August, day: 28 }, { now: new Date('2022-08-28T03:00:00.000Z').getTime() })).toBe(25);
});
`
↑ - Just noticed i forgot to update the test description on this one, my bad

But on this one i only needed to add 2 hours.
`
test('GIVEN now is birthday THEN returns 24', () => {
expect(getAge({ year: 1998, month: Month.February, day: 3 }, { now: new Date('2022-02-03T02:00:00.000Z').getTime() })).toBe(24);
});
`

Thats why i gave up on the method i suggested previously.

lveneris:
Tests passed on my machine and in CI, so I am not sure why they failed for you.

christianfds:
Well, let's go with the adjustments you proposed, if later we notice something odd we can come back here.

Nytelife26:
I suspect that perhaps rather than using `Date.now()` we may need to use the UTC equivalent - that is likely why tests failed on your system until you adjusted for your timezone.

src/lib/birthday/utils.ts
```diff
@@ -104,20 +104,16 @@ export function compareDate(month: Month, day: number, { now = Date.now() }: Tim
...
-	//
-	// * 2021/03/18 - 2020/05/10 = 0
-	// * 2021/03/18 - 2020/02/26 = 1
-	const extra = compareDate(data.month, data.day, { now }) === -1 ? 0 : -1;
-	const years = new Date(now).getUTCFullYear() - data.year;
-	return years + extra;
+export function getAge(date: DateWithOptionalYear, { now = Date.now() }: TimeOptions = {}) {
+	if (date.year === null) return null;
+
+	const birthDate = Date.UTC(date.year, date.month - 1, date.day);
```
christianfds, line 114: Creating it as UTC fixed the issue that was happening.

lveneris:
That's a much cleaner solution, good job.

lveneris (approved):

src/lib/birthday/utils.ts
```diff
@@ -104,20 +104,16 @@ export function compareDate(month: Month, day: number, { now = Date.now() }: Tim
...
+
+	return new Date(now - birthDate).getUTCFullYear() - new Date(0).getUTCFullYear();
```
kyranet, line 116: This is quite the interesting operation, however, we could make it slightly more efficient by not creating a constant over and over. Either create a constant outside this function (and not export), or hardcode `1970` since it's never changing.

Also, it'd be good to same some comment above to show why this works.

As a side note, does this function still work for people born before 1970?

kyranet (changes requested):

src/lib/birthday/utils.ts
christianfds: Yes, it does work for people born before 1970.

I will add some comments explaining the math of it and probably hardcode 1970

lveneris: > As a side note, does this function still work for people born before 1970?

Yes. Since both `now` and `birthDate` are milliseconds since 1970, `(now - birthDate).getUTCFullYear()` is equivalent to `age + 1970`, so we subtract 1970 and it works.

I'm not sure this needs a comment explaining the maths, honestly. The only thing that may need explaining is the aforementioned offset created by the date system. Hardcoding 1970 would be a good idea, though, since I don't envision the UNIX epoch changing in our lifetimes - alternatively, we could simply convert `(now - birthDate)` to a year, but we may lose precision.

Overall, my recommendation would be to create a top-level constant set to `new Date(0).getUTCFullYear()` as it seems more clear than simply hardcoding 1970 and also avoids the constant being recalculated at every call.

UPDATE: No, the compiler will not sufficiently optimize out the call. My recommendation to make a top-level constant, preferably called something like `EPOCH_YEAR`, still stands.

lveneris:
@christianfds is any work still being done on this?

christianfds:
I didn`t have time the last couple days, might do the changes later or maybe tomorrow morning.

kyranet:
Birthday functionality has been removed in the main branch, as per [Skyra v7 plans](https://skyra.notion.site/Skyra-v7-922ba06004654142a7b63347a92513a8).

PR #2192, opened 2022-02-12
feat(crowdin): update localization files

NM-EEA-Y:


kyranet (approved):
This PR deleted more lines than the PR removing the commands and the en-US lines, omg

PR #2207, opened 2022-02-19
refactor: update TLDs

NM-EEA-Y:
*bleep bloop* I updated the TLDs file


kyranet (approved):
At this point ICANN is removing a lot of TLDs and not adding a single one lol

PR #2210, opened 2022-02-21
feat: add dragonite notice to all pokemon commands

favna:
Suggested by @kyranet, lets hope this way Dragonite grows a bit more


kyranet (approved):
I assume it's all tested.

PR #2215, opened 2022-03-01
refactor: remove many deprecated things

kyranet:
Following Skyra v7's plans, we're removing everything we won't support in any of our bots.

- [x] Finish v7 command and hide.
- [x] Write migrations.


favna (changes requested):
1. There is still this interface in `UserRepository.ts` which has 3 references:
```ts
interface RawSpouseEntry {
	user_id_1: string;
	user_id_2: string;
}
```
    Search this file for `spouse` in general.

2. Remove the removed commands from `1594757329224-V13_MigrateAnalytics.json`

3. Add migration to remove rows from `schedule` table where `task_id` is equal to `birthday` or equal to `removeBirthdayRole` (yes the latter exists)

4. In `PermissionNodeManager.test.ts` there is a line:
```ts
const roleSubscriber = createRole({ id: '541743369081192451', name: 'Subscriber', position: 11 }, guild);
```

favna (approved):

kyranet:
yeet

PR #2332, opened 2022-07-27
refactor: v7

kyranet:
This is a heavy WIP PoC of Skyra v7.

*On a side note, migrations are going to be fun to do.*


HigherOrderLogic:
Hope you guy don't get mad for necro-ing this, but is this still WIP? 😅

kyranet:
Well, this is a promising version of Skyra that predates the actual v7 plans, we went for the alternative route of doing the changes incrementally, and make Skyra gradually lighter so she's easier to rewrite.

Current milestone is v6.5, which is "halfway" v6 (keeps discord.js) and v7 (new architecture). I plan to eventually rewrite Skyra into a hybrid of `@skyra/http-framework` with `@discordjs/core`, but there are many challenges around it, which I tried to solve with this PR, but I think this approach is not completely right.

Nevertheless, I want to investigate some things before trying this first, but for now, the migration to Prisma and minimising Skyra's memory are priority atm.

PR #2335, opened 2022-07-30
refactor: remove kitsu and anilist

favna:
| ⚠️ This should not be merged before skyra-project/archid-components#128 and skyra-project/nekokai#51 and merged, and in case of the former, released |
| --- |

As this will now be in nekokai we can remove it from Skyra


kyranet (changes requested):
v7-nekokai needs to be updated to include the anime commands as aliases.

Also, waifu seems to be left, I suppose we'll move it in a following PR? :eyes:

kyranet (approved):

PR #2369, opened 2022-09-11
refactor: remove horoscope/saelem

favna:
This command is largely unused, can be replaced by Ahri's bot,
and we are not interested in maintaining it.
Also because it relies on web scraping it is questionable in legality


src/languages/de-DE/commands/tools.json
```diff
@@ -214,41 +214,7 @@
...
-		"dailyHoroscope": "",
-		"metadata": [
-			"",
-			"",
-			"",
-			""
-		],
-		"metadataTitle": ""
-	},
+
```
kyranet: 🧹 
```suggestion
```

kyranet: And everywhere else too, please, too many files with this.

kyranet (changes requested):

kyranet (approved):

PR #2370, opened 2022-09-11
refactor: simplify regex prefix

kyranet:
Removed `Hey, ` prefix, too Google-y.

Also removed the easter egg `Eva` prefix (based on Skyra's Earth name in the lore) because it was intended to be an easter egg, but some people abused it, so it's not longer one. And because we're switching to slashies someday so those prefixes will be gone sooner or later.


favna (approved):
Psh, I know google responds to "Hey Google" but it's "Ok Google" in my eyes. It IS however always "Hey Siri"

kyranet:
"Ok Boomer" also works.

PR #2387, opened 2022-10-14
feat: shifted to select menu for eval reply options

ikrishagarwal:
This PR adds a feature to the eval command. Before we used to use message-based methods to ask for the desired method of viewing content more than the discord size limit. But now we are using a message action row instead.

Here are some images on how it looks: 
message one
message two

### DIFF
--- a/src/lib/util/Parsers/ExceededLength.ts
+++ b/src/lib/util/Parsers/ExceededLength.ts
@@ -1,12 +1,13 @@
 import { LanguageKeys } from '#lib/i18n/languageKeys';
 import { getHaste } from '#utils/APIs/Hastebin';
-import { promptForMessage } from '#utils/functions';
+import { floatPromise, seconds } from '#utils/common';
+import { deleteMessage } from '#utils/functions';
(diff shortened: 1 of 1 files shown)


src/lib/util/Parsers/ExceededLength.ts
```diff
@@ -32,10 +33,12 @@ export async function handleMessage<ED extends ExtraDataPartial>(
 			await getTypeOutput(message, t, options);
 			return handleMessage(message, options);
 		}
-		case 'haste':
 		case 'hastebin': {
 			if (!options.url) {
-				options.url = await getHaste(options.content ? options.content : options.result!, options.language ?? 'md').catch(() => null);
+				options.url = await getHaste(options.content ? options.content : options.result!, options.language ?? 'md').catch((err) => {
+					container.logger.fatal(err.message);
```
favna: Remove the log, then remove the braces to directly return null again.

favna (changes requested):

favna (approved):

favna:
Tested with Dragonair

kyranet (approved):

PR #2450, opened 2023-04-09
refactor: switch to prisma

kyranet:
For deploying to production, this command **is** required to mark the initial migration as done[^1]:

```sh-session
$ yarn prisma migrate resolve --applied 20240720200159_01_initial
```

[^1]: https://www.prisma.io/docs/orm/prisma-migrate/workflows/baselining


src/.env
```diff
@@ -54,3 +47,7 @@ SENTRY_URL=''
 TOP_GG_TOKEN=''
 TWITCH_CLIENT_ID=''
 TWITCH_TOKEN=''
+
+# The database URL for Prisma, needs to be copied in a local `.env` at the root
+# for Prisma's operations to work.
+DATABASE_URL="postgresql://postgres:postgres@localhost:5432/skyra?schema=public"
```
favna, line 53: Why dont we just put .env in the root altogether and then we can simplify setup from skyra env-utilities as well because it will default to the root.

And for prod we can inline the envs in docker-compose just like we do for the other bots

kyranet: Because then you won't be able to override it with `.env.development.local`, Prisma doesn't do the dotenv-cra behaviour afaik

favna: You're right

favna (approved):

PR #2465, opened 2023-05-28
refactor: start migration to Artiel

kyranet:
> **Warning**: Requires latest works on memes.skyra.pw to be finished before Artiel releases, which is requirement for this PR to merge.


.github/workflows/continuous-delivery.yml
```diff
@@ -41,11 +41,3 @@ jobs:
           tags: |
             skyrabot/skyra:latest
             ghcr.io/skyra-project/skyra:latest
```
favna:
```suggestion
          tags: ghcr.io/skyra-project/skyra:latest
```

If you remove the block below (Dockerhub) then we also shouldn't publish to Dockerhub.

package.json
```diff
@@ -89,7 +86,6 @@
 		"typeorm": "0.2.28",
 		"typeorm-naming-strategies": "^2.0.0",
 		"utf-8-validate": "^6.0.3",
-		"ws": "^8.13.0",
```
favna, line 92: Why `ws` gone? DJS still uses it

src/languages/en-US/commands/admin.json
```diff
@@ -1,5 +1,4 @@
 {
-	"confGuarded": "{{name, toTitleCase}} may not be disabled.",
```
favna, line 2: This is unused and a remnant?

src/languages/en-US/errors.json
```diff
@@ -1,10 +1,9 @@
 {
-	"channelNotReadable": "I am sorry, but I need the permissions **{{VIEW_CHANNEL, permissions}}** and **{{READ_MESSAGE_HISTORY, permissions}}**",
```
favna, line 2: huh?

src/languages/en-US/system.json
```diff
@@ -7,7 +7,6 @@
 	"exceededLengthOutputTime": "{{time}}",
 	"exceededLengthOutputFile": "Sent the result as a file.",
 	"exceededLengthOutputHastebin": "Sent the result to hastebin: {{url}}",
-	"externalServerError": "The external service we use could not process our message. Please try again later.",
```
favna, line 10: ?

src/languages/en-US/system.json
```diff
@@ -29,15 +28,6 @@
...
-	"messagePromptTimeout": "The prompt has timed out.",
-	"noResults": "I wasn't able to find any results for that query",
-	"parseError": "{{REDCROSS}} I failed to process the data I was given, sorry~!",
 	"prefixReminder": "The prefix in this server is set to: `{{prefix}}`",
-	"queryFail": "I am sorry, but the application could not resolve your request. Are you sure you wrote the name correctly?",
-	"textPromptAbortOptions": [
-		"abort",
-		"cancel",
-		"stop"
-	],
```
favna, line 41: all unused?

favna (changes requested):

package.json
kyranet: Remnant from when we had a `ws` handler for the music dashboard. It's a direct dependency of discord.js anyways.

src/languages/en-US/commands/admin.json
kyranet: I removed based on reference count on the `LanguageKeys`'s exports (with some exceptions being Sapphire's keys), and that one... just didn't show up at all.

Same goes for the others.

---

It's true that some commands (including `conf`) are marked as guarded:

https://github.com/skyra-project/skyra/blob/0714191296c43cb95a4a22ace5f116dba226a18b/src/commands/Admin/conf.ts#L19

However, the property is not read (nor it would be a feasible feature considering the type is `commandmatch`, which supports stuff like `Admin` to disallow the entire category):

https://github.com/skyra-project/skyra/blob/0714191296c43cb95a4a22ace5f116dba226a18b/src/lib/database/entities/GuildEntity.ts#L33-L35

Nor the settings system supports filters (like Klasa did for guarded commands):

https://github.com/skyra-project/skyra/blob/0714191296c43cb95a4a22ace5f116dba226a18b/src/lib/database/settings/ConfigurableKey.ts#L11-L44

src/languages/en-US/system.json
kyranet: We don't use `TextPrompt*` anymore, but `ExceededLength*`: https://github.com/skyra-project/skyra/blob/main/src/lib/util/Parsers/ExceededLength.ts

src/languages/en-US/errors.json
kyranet: I have no idea where this comes from, it's not `arguments:` nor `preconditions:` prefixed so it's not Sapphire's, and I could find no references to it yesterday. I have no idea where or when did this line get added, I initially theorized it would be Twitch streams, but it doesn't use it apparently, perhaps because Sapphire handles that?

src/languages/en-US/system.json
kyranet: I also have no idea when or where did this line come from, but as of this PR, Skyra's reliance to external services/APIs are at a minimum. Either way many commands that did implemented its own lines.

favna (dismissed):

src/languages/en-US/commands/admin.json
kyranet: Upon further inspection, I saw that it's actually used, it just has no message:

https://github.com/skyra-project/skyra/blob/4b9d5cdb9fb66c8ea397def042984fdb714fffe1/src/lib/structures/preconditions/PermissionsPrecondition.ts#L47

favna (approved):

favna (approved):

PR #2468, opened 2023-06-06
fix: support new usernames

kyranet:
But also...
- Fixed `reason` command displaying `{{name}}` instead of the parameter.
- Added embed author URLs to message logs where applicable.


favna (dismissed):

favna:
unit tests be hatin

favna (approved):

PR #2478, opened 2023-06-13
refactor: remove `ExceededLength` handler and simplify eval

kyranet:
Also made the character length count check count codepoints instead of characters for extra accuracy.


favna (approved):
Just know that despite this removal I'm never gonna take down `hastebin.skyra.pw` lol. Too useful to have. a stable hastebin.

kyranet:
I'm aware and I wasn't even planning to shut hastebin down, it's still usable outside of the bot, and it's indeed a stable and reliable client.

PR #2505, opened 2023-10-21
refactor: rewrite to djs v14

kyranet:
:eyes:

### Changes
- Upgraded Yarn from v3 to v4.
- ~~Downgraded Node.js CI version from 20 to 18 (and pinned to v18 with Volta) because import loader issues in Workers with Node.js v20. See: https://github.com/privatenumber/tsx/issues/354~~
  - ~~Because the loader in v18 can resolve `.ts` in `Worker`'s constructor, and is required to be passed in the arguments (rather than done at runtime), `scripts/workerTsLoader.mjs` was deleted.~~
- Switched from `jest` to `vitest` for better test performance.
- ~~Switched from `ts-node` to `tsx` for better support in Node.js v18 and upwards (uses `esbuild` under the hood as well).~~ Removed `ts-node` altogether.
- Switched from CJS to ESM.
- Updated `discord.js` from v13 to v14.
  - Updated `discord-api-types/v9` to `discord-api-types/v10` to match.
- Updated `@sapphire/framework` from v2 to v5.
- Updated `typeorm`, giving them a last chance. Skyra now uses less relationships, so it should be *probably fine*.
- Split `SkyraCommand` (extends `Subcommand`) into `SkyraCommand` (extends `Command`) and `SkyraSubcommand` (extends `Subcommand`).
- Removed `PaginatedMessageCommand`, replaced with `Skyra{Subc,C}ommand.PaginatedOptions` to reduce code duplication.
- Removed old `v7-*` commands. Kept `v7-artiel` since it's newer and expected to be used soon.
- Removed unused DB structures (`banners` and `client`).
- Removed left-over API route from the social module (`/banners`).
- Removed `TriviaManager` tests, they're being removed soon in favour of Artiel.

### Extras
- Updated `s!conf`'s menu to listen only to reactions on the menu's message
- Updated `s!dice` to default parameter-less to `6`
- Updated `getImage` to fallback to stickers when there are no attachments
- Updated `channelName` argument to throw a less misleading error when filter returns false


.github/workflows/continuous-integration.yml
favna, line 1: As per the conversation on Discord, downgrading Node is not a solution. We will not merge this PR until it fully supports Node 20.

src/.env.development
```diff
@@ -9,8 +9,6 @@ SHARDS='"auto"'
 CLIENT_PRESENCE_NAME='sd!help'
 CLIENT_PRESENCE_TYPE='LISTENING'
 
-SISTER_CLIENTS='338249781594030090 419828209966776330 444081201297227776'
```
favna, line 12: awe Skyra's sisters are dead

src/arguments/snowflake.ts
```diff
@@ -1,7 +1,7 @@
 import { LanguageKeys } from '#lib/i18n/languageKeys';
-import { Argument, ArgumentContext } from '@sapphire/framework';
+import { Argument } from '@sapphire/framework';
 import { DiscordSnowflake } from '@sapphire/snowflake';
-import type { Snowflake } from 'discord.js';
+import type { Snowflake } from 'discord-api-types/v10';
```
favna: Import from discord.js instead. DJS re-exports all of dapi-types v10.

src/commands/Admin/conf.ts
```diff
@@ -1,33 +1,40 @@
...
+import { PermissionLevels } from '#lib/types';
 import { isValidCustomEmoji, isValidSerializedTwemoji, isValidTwemoji } from '#lib/util/functions/emojis';
 import { filter, map } from '#utils/common';
 import { inlineCode } from '@discordjs/builders';
 import { ApplyOptions, RequiresClientPermissions } from '@sapphire/decorators';
 import { CommandOptionsRunTypeEnum } from '@sapphire/framework';
 import { send } from '@sapphire/plugin-editable-commands';
 import { toTitleCase } from '@sapphire/utilities';
-import { PermissionFlagsBits } from 'discord-api-types/v9';
+import { PermissionFlagsBits } from 'discord-api-types/v10';
```
favna: import from discord.js, same goes for imports from @discordjs subpackage, all should route through discord.js. Not gonna repeat it after this comment but it goes for all imports.

src/commands/Admin/conf.ts
```diff
@@ -1,33 +1,40 @@
...
+@ApplyOptions<SkyraSubcommand.Options>({
 	aliases: ['settings', 'config', 'configs', 'configuration'],
 	description: LanguageKeys.Commands.Admin.ConfDescription,
 	detailedDescription: LanguageKeys.Commands.Admin.ConfExtended,
 	guarded: true,
 	permissionLevel: PermissionLevels.Administrator,
 	runIn: [CommandOptionsRunTypeEnum.GuildAny],
-	subCommands: ['set', { input: 'add', output: 'set' }, 'show', 'remove', 'reset', { input: 'menu', default: true }]
+	subcommands: [
+		{ name: 'set' },
```
favna: where is the `messageRun` for this and the others lacking it?

src/commands/Management/Message Filters/attachmentMode.ts
```diff
@@ -1,8 +1,7 @@
-import { AdderKey, GuildEntity, GuildSettings } from '#lib/database';
+import { GuildSettings, type AdderKey, type GuildSettingsOfType } from '#lib/database';
```
favna, line 1: Awe no more [Adder](https://en.wikipedia.org/wiki/Adder)[Key](https://en.wikipedia.org/wiki/Key)

src/commands/Moderation/Management/history.ts
```diff
@@ -101,16 +103,15 @@ export class UserCommand extends SkyraCommand {
...
+					embed.addFields(this.displayModerationLogFromModerators(usernames, now, entry));
 				}
 
 				return embed;
 			});
 		}
 
 		await display.run(response, message.author);
-		return response;
+		await response;
```
favna: > 'await' has no effect on the type of this expression. ts(80007)

src/lib/database/entities/GuildEntity.ts
```diff
@@ -670,7 +672,7 @@ export class GuildEntity extends BaseEntity implements IBaseEntity {
 	 * Gets the [[Language]] for this entity.
 	 */
 	public getLanguage(): TFunction {
-		return container.i18n.getT(this.language);
+		return getT(this.language as LocaleString);
```
favna: Why no longer `container.i18n.getT`? They should be identical

src/lib/database/settings/structures/AdderManager.ts
favna, line 1: Dang we do have [Adder](https://en.wikipedia.org/wiki/Adder)s after all LOL

src/lib/i18n/translate.ts
```diff
@@ -86,3 +114,31 @@ export type TResolvable = SkyraArgs | TFunction;
...
+export async function fetchLanguage(context: InternationalizationContext) {
+	return ((await container.i18n.fetchLanguage(context)) ?? 'en-US') as LocaleString;
+}
+
+/**
+ * Fetches the translation function for the given context.
+ * @param context The internationalization context.
+ * @returns The translation function.
+ */
+export async function fetchT(context: InternationalizationContext) {
```
favna, line 142: We should use `fetchT` from the plugin, not this custom function.

src/lib/i18n/translate.ts
```diff
@@ -86,3 +114,31 @@ export type TResolvable = SkyraArgs | TFunction;
...
+	return container.i18n.getT(locale ?? 'en-US');
+}
+
+/**
+ * Fetches the language for the given {@link InternationalizationContext}.
+ * If the language cannot be fetched, defaults to 'en-US'.
+ * @param context The InternationalizationContext to fetch the language for.
+ * @returns The fetched language as a {@link LocaleString}.
+ */
+export async function fetchLanguage(context: InternationalizationContext) {
```
favna, line 133: We should use `fetchLangauge` from the plugin, not this custom function. It already has option to default to `en-US`, implement `fetchLanguage` on the client options and send the default if none of the context options can send another language.

src/lib/i18n/translate.ts
```diff
@@ -86,3 +114,31 @@ export type TResolvable = SkyraArgs | TFunction;
 export function resolveT(t: TResolvable): TFunction {
 	return typeof t === 'function' ? t : t.t;
 }
+
+/**
+ * Returns a translation function for the specified locale, or the default 'en-US' if none is provided.
+ * @param locale The locale to get the translation function for.
+ * @returns The translation function for the specified locale.
+ */
+export function getT(locale?: LocaleString | Nullish) {
```
favna, line 123: We should be able to use the `container.i18n.getT` instead of this function, see about default language above.

src/lib/util/common/objects.ts
```diff
@@ -0,0 +1,5 @@
+export function omit<T extends object, K extends keyof T>(object: T, ...keys: readonly K[]): Omit<T, K> {
```
favna: should really port this to @sapphire/utilities tbh, and then we should name it `omitKeyFromObject`

src/routes/guilds/guild/settings.ts
```diff
@@ -70,7 +78,7 @@ export class UserRoute extends Route {
 			// If null is passed, reset to default:
 			if (value === null) return [entry.property, entry.default];
 
-			const ctx = { ...context, entry };
+			const ctx = { ...context, entry } as SerializerUpdateContext;
```
favna: `as Serializer.UpdateContext`

tests/lib/database/settings/structures/PermissionNodeManager.test.ts
```diff
@@ -19,7 +19,7 @@ describe('PermissionNodeManager', () => {
 		return entity.permissionNodes['sorted'];
 	}
 
-	describe('run', () => {
+	describe.skip('run', () => {
```
favna: TODO

tests/lib/database/settings/structures/PermissionNodeManager.test.ts
```diff
@@ -105,20 +105,20 @@ describe('PermissionNodeManager', () => {
 		});
 	});
 
-	describe('remove', () => {
-		describe('user', () => {
+	describe.skip('remove', () => {
```
favna: TODO

tests/mocks/MockInstances.ts
```diff
@@ -147,7 +158,11 @@ function addCommand(command: SkyraCommand) {
 	for (const alias of command.aliases) commands.aliases.set(alias, command);
 }
 
-class Command extends SkyraCommand {}
+class Command extends SkyraCommand {
+	public override messageRun(): unknown {
```
favna: Implement this as an arrow function then use `vi.fn()`. Should be `public override messageRun = vi.fn()` I think.

tests/tsconfig.json
```diff
@@ -12,7 +12,10 @@
 			"#root/*": ["../src/*"]
 		}
 	},
-	"include": [".", "../src/**/*"],
-	"exclude": ["../dist/**/*"],
-	"references": [{ "path": "../src" }]
+	"include": [".", "../src"],
+	"references": [{ "path": "../src" }],
+	"ts-node": {
```
favna: ts-node? I'm barfing. Please no.

vitest.config.ts
```diff
@@ -0,0 +1,93 @@
+import { resolve } from 'node:path';
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+	resolve: {
+		alias: [
```
favna, line 6: Is this actually needed? I don't need it for dragonite for example

favna (commented):
I may have made some mistakes in reviewing because there were so many files

favna (changes requested):
Now with request changes

tests/lib/database/settings/structures/PermissionNodeManager.test.ts
kyranet: Those aren't even going to be used with slash commands. I'm skipping those because it's wasting a lot of time and energy, specially considering the system has proven itself from the +2 years it's been implemented and running in production.

tests/lib/util/functions/emojis/getEmojiObject.test.ts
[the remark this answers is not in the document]
kyranet: It's in the `tsconfig.json`, but TS refuses to acknowledge that fact.

vitest.config.ts
kyranet: It somehow didn't work at all without, it's a first for me too, also incredibly annoying, but I'm out of ideas.

kyranet: Tried removing it, it didn't work because it'll try to use the built JS files rather than the source files.

Code_s8cDcxFjAH

*I mean, if we require building before testing, we solve this issue.*

src/lib/i18n/translate.ts
kyranet: `fetchT` doesn't exist in the plugin:
image

src/lib/i18n/translate.ts
kyranet: The plugin's `fetchLanguage` doesn't return `LocaleString` but rather `string`. This extra strictness helps making the code more resilient to typos. Also, unlike the plugin's `fetchLanguage`, this function is typed to never return `null`.

src/lib/i18n/translate.ts
kyranet: Same as previous comment, this function has stricter types, requiring `LocaleString` rather than just any `string`. Also provides a default so `getT()` is valid and gets the default language.

vitest.config.ts
favna: nah lets keep aliases then

src/lib/util/common/objects.ts
favna: https://github.com/sapphiredev/utilities/pull/700

src/lib/moderation/managers/ModerationManager.ts
```diff
@@ -173,7 +171,7 @@ export class ModerationManager extends Collection<number, ModerationEntity> {
 			})
 		);
 
-		return lock.resolve;
+		return () => lock.resolve();
```
favna, line 174: can't we use `lazy` from @sapphire/utilities here? Or will that be an issue because it caches?
```ts
/**
 * Lazily creates a constant or load a module and caches it internally
 * @param cb The callback to lazily run
 * @returns The value returned by the callback, or the cached value if it was already initialised once.
 */
export function lazy<T>(cb: () => T) {
	let defaultValue: T;

	return () => (defaultValue ??= cb());
}

```

tests/lib/database/settings/structures/PermissionNodeManager.test.ts
favna: Then delete the code

tests/lib/database/settings/structures/PermissionNodeManager.test.ts
favna: Ditto, delete

tsconfig.base.json
```diff
@@ -1,6 +1,8 @@
 {
-	"extends": "@sapphire/ts-config",
+	"extends": ["@sapphire/ts-config", "@sapphire/ts-config/extra-strict", "@sapphire/ts-config/decorators", "@sapphire/ts-config/verbatim"],
 	"compilerOptions": {
+		"module": "NodeNext",
```
favna: We should be able to remove this. NodeNext is identical to Node16 which is used by sapphire by default.

favna (changes requested):
Couple of comments

src/lib/moderation/managers/ModerationManager.ts
kyranet: That's not what the code does, it registers a lock, and the callback unlocks and unregisters it.

favna (approved):

favna (approved):
LGTM

PR #2540, opened 2024-01-03
refactor(i18next): use `formatters`

kyranet:


favna (commented):
why are not all formatters cached?

kyranet:
@favna cached formatters return a callback and are called only once per `lng` and `options`, which is useful for shared instances. There's nothing to cache otherwise.

favna (approved):

PR #2553, opened 2024-01-26
refactor: improved `whois` with interaction support

kyranet:
Implements the base for future interaction-based commands in Skyra, starting with `whois`.

<table>
<tr>
 <td>Old
 <td>New
<tr>
 <td>

image

 <td>

image

</table>

As shown above, the response's format has also been tweaked for cleanliness.

The relevant debug logs are added too with this PR, distinguishing from the command type:
image

Furthermore, a deprecation message is sent linking to the slash command when a message command is used:
image

And lastly, the new i18n files are to slowly accomodate Skyra to the format used in the HTTP bots, slowly migrating her to the same design.

### DIFF
--- a/src/commands/Tools/whois.ts
+++ b/src/commands/Tools/whois.ts
@@ -1,23 +1,47 @@
 import { SkyraEmbed } from '#lib/discord';
 import { LanguageKeys } from '#lib/i18n/languageKeys';
+import { getSupportedUserLanguageT } from '#lib/i18n/translate';
 import { SkyraCommand } from '#lib/structures';
 import type { GuildMessage } from '#lib/types';
 import { PermissionsBits } from '#utils/bits';
-import { map, months, seconds } from '#utils/common';
-import { Colors, Emojis } from '#utils/constants';
+import { desc, map, maybeParseDate, months, seconds } from '#utils/common';
+import { Colors, EmojiData, Emojis } from '#utils/constants';
 import { getDisplayAvatar, getTag } from '#utils/util';
-import { TimestampStyles, time } from '@discordjs/builders';
+import { ActionRowBuilder, ButtonBuilder, TimestampStyles, time } from '@discordjs/builders';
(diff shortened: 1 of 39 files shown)


favna (approved):

src/commands/Tools/whois.ts
```diff
@@ -40,98 +64,145 @@ export class UserCommand extends SkyraCommand {
...
+				.addUserOption((option) => applyLocalizedBuilder(option, Root.User).setRequired(true))
+				.setDMPermission(false)
+		);
+
+		registry.registerContextMenuCommand((builder) =>
+			applyNameLocalizedBuilder(builder, Root.ContextMenuName) //
+				.setType(ApplicationCommandType.User)
+				.setDMPermission(false)
+		);
+	}
```
favna, line 107: No `idHints` defined and bulk overwrite isnt set as the default yet.

favna (changes requested):

src/commands/Tools/whois.ts
favna: `src/lib/setup.ts`
```ts
// Config must be the first to be loaded, as it sets the env:
import '#root/config';
import 'reflect-metadata';

// Then we import the other side-effect loaders:
import '#lib/setup/PaginatedMessage';
import '#utils/Sanitizer/initClean';
import '@sapphire/plugin-api/register';
import '@sapphire/plugin-editable-commands/register';
import '@sapphire/plugin-i18next/register';
import '@sapphire/plugin-logger/register';

// And finally the modules where we want to execute some code for:
import { ApplicationCommandRegistries, RegisterBehavior } from '@sapphire/framework';
import * as colorette from 'colorette';
import { inspect } from 'node:util';

inspect.defaultOptions.depth = 1;
colorette.createColors({ useColor: true });
ApplicationCommandRegistries.setDefaultBehaviorWhenNotIdentical(RegisterBehavior.BulkOverwrite);
```

favna (approved):

PR #2569, opened 2024-02-09
refactor(moderation): separate `type` and `metadata`

### LINKED ISSUE #1335: rewrite: moderation module improvements

kyranet:
This is a large change, it breaks the mix between a moderation case's type and its metadata.

For years, Skyra stored a moderation case's type and metadata into a single field of 4 bits each. This was done so the data would be stored far more efficiently, using a single smallint rather than an integer and multiple booleans.

The structure would be
```
XXXXYYYY
```
Where `XXXX` is the metadata (a bitfield) and `YYYY` the type (an enum).

However, while this system was made when we had less moderation types, we quickly filled all of them with the restriction actions, and they cannot be removed as we need to preserve backwards compatibility. This wasn't an issue back then...

... until years later Discord added timeouts, a native way to mute users, and we couldn't support them because the type bits were all used up.

This PR fixes this issue, it extracts both fields into their separate entries, this also simplifies some code by a lot, but since a lot of logic changed, careful testing needs to be done to check everything is fine.

*And hopefully some day in the near future, Skyra will support timeouts.*


favna (approved):
Approved assuming you tested the database migration

PR #2577, opened 2024-02-10
build: better docker and typeorm config

favna:


ormconfig.js
```diff
@@ -1 +1,8 @@
-export default (await import('./dist/lib/database/database.config.js')).AppDataConfig;
+import { AppDataConfig } from '#lib/database/database.config';
```
kyranet, line 1: This should probably be `import()`'d after `setup()` so the environment variables are loaded

kyranet (approved):

PR #2594, opened 2024-03-14
feat: add timeout support

kyranet:
## Changelog

- Add `s!timeout` and `s!untimeout`.
- Add `events.timeout` configuration key, for logging manual actions.
- Manual or external prunes now report the moderator more consistently.

### Private changelog

#2595 introduced a few bugs... all fixed in this PR.

- Added action hooks for the moderation logger.
- Fixed double completion, causing timed entries to duplicate the logs.
- Fixed `ModerationActionTimeout` not setting the data correctly.

PR #2595, opened 2024-03-15
refactor(Moderation): rewrite backend and commands

### LINKED ISSUE #1335: rewrite: moderation module improvements

### LINKED ISSUE #1833: request: streamline viewing moderation logs for a specific user

kyranet:
> [!IMPORTANT]
> ~~This PR needs extensive testing, as the behaviour of many parts were rewritten due to the Actions layer being the highest level API the entire moderation front-end uses.~~

This PR rewrites the Actions layer in the moderation module, changing the functional style `ModerationActions` class which had a lot of code duplication and almost 1000 lines of code, to simple classes for each action that define what each action does in a comprehensive way.

## Changelog

- Added generic error handling for chat input commands.
- Added `timeout` count in `/case list overview:True`.
- Improved the style in `/case show` and moderation logs.
- Improved the style in `/case list`.
- Improved action status checking.
- Improved ban check in `s!unban`.
- Deprecated `s!history` for `/case list`.
- Deprecated `s!moderations` for `/case list`.
- Deprecated `s!mutes` for `/case list type:Mute`.
- Deprecated `s!warnings` for `/case list type:Warnings`.
- Deprecated `s!reason` for `/case edit reason:<>`.
- Deprecated `s!time` for `/case edit duration:<>`.
- Deprecated `s!unwarn` for `/case archive`.
- Fixed unmutes not giving the roles back.
- Fixed moderation actions not cancelling pending schedules correctly.
- Fixed `Set Nickname`, `Role Add`, and `Role Remove` actions not undoing correctly if re-scheduled.
- Fixed moderation logs not being localized.
- Fixed `es-419` not falling back to `en-US` when a key is not available in `es-ES`.
- Fixed the message for upper ranges not being inclusive.

---

For years, I received one-off issues with Skyra's moderation module, specially the mute system, because it wasn't giving the roles back. And alongside with the eventual implementation of timeouts, #2594 got way messier than it should have been.

As such, I decided to rewrite the Actions layer. This is the 3rd layer and is the heart of the moderation module, where the actions are applied at and where the database is updated.

*(btw, there's a timeout action now, the aforementioned PR will be updated to add the command and the necessary events)*

With the new Actions layer, I wanted to reconsider the moderation entry metadata and give them more meaning:
- `Appeal` became `Undo`, which better conveys the meaning of the type.
- `Fast` was originally intended for fast moderation actions (under X time such as 15 seconds) that wouldn't be logged in the moderation logs channel. However, that's both a maintenance headache and an extremely niche feature nobody asked for. This bit is phased out and deprecated.
- `Invalidated` became `Archived`, which finally clears what it means, and in a backwards compatible way.

Because `Archived`, I finally realised that any moderation entry should be archivable, not just warnings, so I removed the `unwarn` command and added a new subcommand in the `case` command. This was a bit of a scope creep, but I felt like it was a very needed change. I deprecated the messageRun handlers in the command and slashified it entirely, allowing me to merge 7 commands in a single one.

And with the new command, case deletes are now a more obvious posibility, which caught the attention and got more testing. That led to the discovery of new edge cases.

So because of these edge cases, I'm also rewriting the Data layer (4th) and stripping TypeORM entirely from the layer's API using a class compatible with both TypeORM and Prisma. This new layer has stricter types and is more accurate.

---

A funny tale; some users know that Skyra's unmute was unreliable *for years*, this is an issue I couldn't figure out and mostly blamed at Discord because it happens sometimes, but not always.

When I rewrote the Actions layer, I found a few issues.

Skyra was looking for the cases as intended, and cancelling the last entry. However...
- 🚩 it was looking for **non-temporary** actions, so temporary mutes would be ignored.

And then when I found out I had to rewrite the Data layer...
- 🚩 "last" doesn't mean "most recent", not with a `Collection`, which is insert-sorted.

---

*As a side note, this PR also backports several changes from #2593.*


kyranet:
Marking as draft because there's a lot more work I'd like to do before it's ready.

- [x] Clean up redundant features.
  - [x] Start using `getTranslationKey` and remove duplicated i18n keys.
- [x] Handle rescheduling in moderation entry update.
- [x] Fix `/case list`.

kyranet:
Ok, I realised there's one more thing left to do, QoL UX. When deleting a previous task, mark the linked case as its timer has been cancelled, updating the log accordingly and blocking duration updates in `/case edit`.

Other than that, I believe it's ready.

- [x] `TypeMetadata.Completed`.
  - [x] Implement special scheduling behaviour.
  - [x] Implement automatic task deletion on set.
  - [x] Block `/case edit time:<>` on completed tasks.
- [x] Muted users can be muted again.
  - [x] **UX** Define action-specific keys to make the error message clearer.
- [x] Option in `/case list` to list only pending items (hasn't expired or hasn't been aborted).
- [x] Update `LanguageKeys.Commands.Moderation` references in `ModerationCommand`.

favna:
Does this also fix #1833?

kyranet:
As a matter of a fact, yes, it does. Linked the issue now, ty

src/commands/Moderation/Utilities/case-deprecations.ts
```diff
@@ -0,0 +1,47 @@
+import { LanguageKeys } from '#lib/i18n/languageKeys';
+import { SkyraCommand } from '#lib/structures';
+import { PermissionLevels, type GuildMessage } from '#lib/types';
+import { ApplyOptions } from '@sapphire/decorators';
+import { CommandOptionsRunTypeEnum } from '@sapphire/framework';
+import { send } from '@sapphire/plugin-editable-commands';
+import { chatInputApplicationCommandMention } from 'discord.js';
+
+@ApplyOptions<SkyraCommand.Options>({
+	name: '\u200Bcase-deprecations',
```
favna, line 10: why the zero-width space? Also i'm pretty sure we export a zws constant somewhere

src/languages/en-US/errors.json
```diff
@@ -1,4 +1,13 @@
 {
+	"genericUnknownChannel": "I'm sorry, I tried to operate on a channel unknown to Discord and failed, this error has been reported to the developers.",
+	"genericUnknownGuild": "I'm sorry, I tried to operate on a server unknown to Discord and failed, this error has been reported to the developers.",
+	"genericUnknownMember": "I'm sorry, I tried to operate on a member unknown to Discord and failed, this error has been reported to the developers.",
+	"genericUnknownMessage": "I'm sorry, I tried to operate on a message unknown to Discord and failed, this error has been reported to the developers.",
+	"genericUnknownRole": "I'm sorry, I tried to operate on a role unknown to Discord and failed, this error has been reported to the developers.",
+	"genericMissingAccess": "I'm sorry, I tried to operate on a resource without the necessary permissions and failed, this error has been reported to the developers.",
```
favna:
```suggestion
	"genericUnknownChannel": "I'm sorry, I tried to perform an action for a channel unknown to Discord and failed, this error has been reported to the developers.",
	"genericUnknownGuild": "I'm sorry, I tried to perform an action for a server unknown to Discord and failed, this error has been reported to the developers.",
	"genericUnknownMember": "I'm sorry, I tried to perform an action for a member unknown to Discord and failed, this error has been reported to the developers.",
	"genericUnknownMessage": "I'm sorry, I tried to perform an action for a message unknown to Discord and failed, this error has been reported to the developers.",
	"genericUnknownRole": "I'm sorry, I tried to perform an action for a role unknown to Discord and failed, this error has been reported to the developers.",
	"genericMissingAccess": "I'm sorry, I tried to perform an action for a resource without the necessary permissions and failed, this error has been reported to the developers.",
```

While `operate on` is technically correct, it's more commonly related to surgical operations.

src/lib/structures/data/SortedCollection.ts
favna, line 1: Wouldn't it be better for this to extends DiscordJS Collection? could it require less code that way?

src/lib/util/constants.ts
```diff
@@ -38,6 +38,12 @@ export const enum Emojis {
 	GreenTickSerialized = 's637706251253317669',
 	Loading = '<a:sloading:656988867403972629>',
 	RedCross = '<:redCross:637706251257511973>',
+	Calendar = '<:calendar_icon:1218607529702068294>',
+	SandsOfTime = '<:sandsoftime:1218608481565802587>',
```
favna: That's called an hourglass, not a sands of time. Is that what it's called in Spanish? Y'all nuts if so.

Sands of time is a [Prince of Persia game](https://en.wikipedia.org/wiki/Prince_of_Persia:_The_Sands_of_Time) with [some banger tunes](https://youtu.be/naPhZbhDxMQ)

src/listeners/commands/_chat-input-shared.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export async function handleCommandError(error: unknown, payload: ChatInputCommandErrorPayload | ChatInputSubcommandErrorPayload) {
+	const { interaction } = payload;
+	const t = getSupportedUserLanguageT(interaction);
+	const resolved = flattenError(payload.command, error);
+	const content = resolved ? resolveError(t, resolved) : generateUnexpectedErrorMessage(interaction.user.id, payload.command, t, error);
+
+	try {
+		if (interaction.replied) await interaction.followUp({ content, flags: MessageFlags.Ephemeral });
+		else if (interaction.deferred) await interaction.editReply({ content });
```
favna: Why not call `editReply` on `replied` as well? That's what I do for Dragonite also.

src/listeners/commands/_chat-input-shared.ts
```diff
@@ -0,0 +1,18 @@
...
+		else await interaction.reply({ content, flags: MessageFlags.Ephemeral });
```
favna:
```suggestion
		else await interaction.reply({ content, ephmemeral: true });
```
Why use flags and not just `ephmemeral: true`?

src/listeners/commands/_chat-input-shared.ts
```diff
@@ -0,0 +1,18 @@
...
+import { flattenError, generateUnexpectedErrorMessage, resolveError } from './_shared.js';
+
+export async function handleCommandError(error: unknown, payload: ChatInputCommandErrorPayload | ChatInputSubcommandErrorPayload) {
+	const { interaction } = payload;
+	const t = getSupportedUserLanguageT(interaction);
+	const resolved = flattenError(payload.command, error);
+	const content = resolved ? resolveError(t, resolved) : generateUnexpectedErrorMessage(interaction.user.id, payload.command, t, error);
+
+	try {
+		if (interaction.replied) await interaction.followUp({ content, flags: MessageFlags.Ephemeral });
```
favna:
```suggestion
		if (interaction.replied) await interaction.followUp({ content, ephmemeral: true });
```

src/listeners/commands/_chat-input-shared.ts
```diff
@@ -0,0 +1,18 @@
+import { getSupportedUserLanguageT } from '#lib/i18n/translate';
+import { type ChatInputCommandErrorPayload } from '@sapphire/framework';
+import type { ChatInputSubcommandErrorPayload } from '@sapphire/plugin-subcommands';
+import { MessageFlags } from 'discord.js';
+import { flattenError, generateUnexpectedErrorMessage, resolveError } from './_shared.js';
```
favna:
```suggestion
import type { ChatInputSubcommandErrorPayload } from '@sapphire/plugin-subcommands';
import { flattenError, generateUnexpectedErrorMessage, resolveError } from './_shared.js';
```

If the other 2 are accepted

src/config.ts
```diff
@@ -194,7 +194,7 @@ function parseInternationalizationOptions(): InternationalizationOptions {
 			load: 'all',
 			lng: 'en-US',
 			fallbackLng: {
-				'es-419': ['es-ES'], // Latin America Spanish falls back to Spain Spanish
+				'es-419': ['es-ES', 'en-US'], // Latin America Spanish falls back to Spain Spanish
```
favna, line 197: Why is en-US here? In fact, I thought @vladfrangu updated discord-api-types recently to support es-419?

favna (changes requested):
Couple of things

src/lib/structures/data/SortedCollection.ts
kyranet: Can’t, Collection extends Map, and I don’t have access to its insertion order nor is its data structure desirable for this purpose.

src/listeners/commands/_chat-input-shared.ts
kyranet: ~~In case the reply contains a meaningful message, but I can edit it to do editReply, yeah.~~ To ensure a reply is sent ephemeral.

src/listeners/commands/_chat-input-shared.ts
kyranet: ephemeral*, and used to because the HTTP bots, but I can update them.

src/config.ts
kyranet: Because if a key doesn’t exist in es-419, it’ll fallback to es-ES.

And if it also doesn’t exist in es-ES, without en-US, i18next won’t be able to resolve the key.

This is merely so untranslated keys fallback correctly. It has nothing to do with Discord, by the way.

src/commands/Moderation/Utilities/case-deprecations.ts
kyranet: We haven't used them for the v7 deprecation commands, so I haven't used them either for this.

favna (approved):

PR #2615, opened 2024-04-07
feat: remove love, pop, rate

OfficialSirH:
Completes https://github.com/skyra-project/artiel/pull/133


kyranet (changes requested):
You have to delete the strings from `en-US/commands/fun.json` too

kyranet (approved):
Sneaky `escaperopeOutput`, thankfully you noticed that in between, because I would have probably deleted it with the rest LOL

favna (approved):

kyranet (approved):

PR #2631, opened 2024-06-03
fix: use discord.js-utilities functions

favna:


docker-compose.yml
```diff
@@ -17,7 +17,7 @@ services:
         max-file: '3'
 
   postgres:
-    container_name: postgres
```
kyranet: What is this change?

kyranet (changes requested):

docker-compose.yml
```diff
@@ -17,7 +17,7 @@ services:
...
+    container_name: postgres2
```
favna:
```suggestion
    container_name: postgres
```

docker-compose.yml
```diff
@@ -60,7 +60,7 @@ services:
     container_name: skyra
     depends_on:
       - influx
-      - postgres
+      - postgres2
```
favna:
```suggestion
      - postgres
```

docker-compose.yml
favna: Something I pushed accidentally

kyranet (approved):

PR #2659, opened 2024-08-17
refactor(settings): replace RWLock for write lock

kyranet:


kyranet:
It seems I *do* need a lock, not a RwLock, but a write lock, to make sure I don't do two updates (or two inserts) simultaneously.

favna (approved):

PR #2684, opened 2024-10-19
fix: fixes for latest djs update

favna:


src/listeners/twitch/twitchStreamOffline.ts
```diff
@@ -43,7 +43,7 @@ export class UserListener extends Listener<Events.TwitchStreamOffline> {
 
 				// Retrieve the channel to send the message to
 				const channel = guild.channels.cache.get(guildSubscription.channelId) as TextBasedChannelTypes;
-				if (isNullish(channel) || !canSendMessages(channel)) {
+				if (isNullish(channel) || !canSendMessages(channel) || channel.isDMBased()) {
```
kyranet, line 46: This should be casted to a guild text based channel instead.

tsconfig.base.json
```diff
@@ -1,6 +1,7 @@
 {
 	"extends": ["@sapphire/ts-config", "@sapphire/ts-config/extra-strict", "@sapphire/ts-config/decorators", "@sapphire/ts-config/verbatim"],
 	"compilerOptions": {
-		"removeComments": true
+		"removeComments": true,
+		"skipLibCheck": true
```
kyranet, line 5: Any reason for this change?

package.json
```diff
@@ -114,13 +114,12 @@
...
 		"tsc-watch": "^6.2.0",
-		"typescript": "^5.4.5",
+		"typescript": "~5.4.5",
 		"vitest": "^2.1.3"
 	},
 	"resolutions": {
-		"@types/node": "20.11.5",
 		"ansi-regex": "^5.0.1",
-		"discord-api-types": "^0.37.98",
+		"discord-api-types": "0.37.97",
```
kyranet, line 122: I'm pretty sure latest djs is on latest -types (and with ^), unless that's still not released cc: @vladfrangu

kyranet (changes requested):

tsconfig.base.json
favna: error in /decorators where MethodDecorator is somehow no longer generic

package.json
favna: I checked what's in yarn.lock... :\

kyranet:
Superseded by #2668 with the fixes I wanted to have

PR #2733, opened 2025-02-26
feat: add `enforceNonce: true`

kyranet:
https://github.com/discordjs/discord.js/blob/bb6767113fb738e4ca7e9fdfb853bd4fa091c02a/packages/discord.js/src/structures/MessagePayload.js#L124-L129


favna (approved):
what's a nonce other than british slang curse word?

kyranet:
A random value sent to verify if a message went through just fine. In this case, it makes it so Discord deduplicates requests with the same `nonce` value

